PR #116, opened 2021-07-07
📝 Updated Quick Start example

iammordaty:
Hi, I noticed that passing incoming directory is required by configuration:

```
$ docker run -d \
  -p 5000:5000 \
  -p 5001:5001 \
  -v <redacted>:/var/slskd/downloads \
  -v <redacted>:/var/slskd/shared \
  -e "SLSKD_SLSK_USERNAME=<redacted>" \
  -e "SLSKD_SLSK_PASSWORD=<redacted>" \
  --name slskd \
  slskd/slskd:latest

$ docker logs slskd
Invalid configuration:
  Directories:
    The Incomplete field specifies a non-existent directory '/var/slskd/incomplete'.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ docker run -d \
   -p 5000:5000 \
   -p 5001:5001 \
   -v <path/to/downloads>:/var/slskd/downloads \
+  -v <path/to/incoming>:/var/slskd/incoming \
   -v <path/to/shared>:/var/slskd/shared \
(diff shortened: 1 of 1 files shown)


jpdillingham:
Thanks!  This is actually a bug with the current Dockerfile (these paths all default to directories within the container in the `canary` tag), but most people will need to set these paths regardless.

Issue #97, opened 2021-06-13
Split runtime configuration from user preferences, save user preferences in a database


jpdillingham:
Presently application and Soulseek options are implemented through the `Options` class and populated via a yaml file, environment variables, or command line arguments.  This works well for steady-state configuration things that impact how the application starts and runs, and for things that won't change often (if ever), but not well at all for things that a user might change after the application is running.

The current `Options` class should be renamed to `Configuration` (or similar) to differentiate the contents from user options or preferences set at runtime.

A new class that is similar in structure to the current `Options` class named `Preferences` (or similar) should be created to represent user options.

To keep this work a manageable size, `Preferences` should be empty, but the application should be able to save and load preferences from disk at startup and when preferences are changed by the user.  Followup issues will be created to track the work to port existing things from configuration to preferences, starting with Soulseek login credentials and network proxy configuration.

jpdillingham:
After putting a ton of thought and experimentation into this, I've decided that the existing yaml file will be used for runtime configuration.  The app will (at some later point) update this file with changes from the UI, and internals will react to changes.  This leverages the .NET "options pattern", has reactivity and snapshotting baked in, and solves for cases where a user might manually update configuration themselves (or with some external app).

This introduces some complexity in the code, as option consumers now need to carefully consider which options they are consuming and whether they'll react to or respect updates made after the application starts.  I've included detailed documentation in the `Options` class to clarify usage.

There will be options that don't lend themselves to specification via environment variables or the command line, and users will be forced to use the configuration file for those.

I've created #124 to allow option lists to be specified via the command line or environment variables, but complex objects are not going to be supported until there's a really strong use case.

PR #121, opened 2021-07-12
Don't require Soulseek credentials, don't automatically connect if they are missing

### LINKED ISSUE #119: Make connection to Soulseek server on start optional
Currently the service creates a client and immediately connects on start.  To make way for configuration changes, refactor the service so that it won't start automatically unless all options are in place and a new option to enable automatic connecting is enabled.

Allow the service to be started manually as a result of a user request, or automatically as options are specified and/or the option to enable automatic connecting changes from disabled to enabled.

The automatic start may be a bit tricky; it will require monitoring options for updates (for example, if the yaml file is updated).

There's a lot of moving parts to this:

Upon start:
* [x] If the username and password fields are specified, connect immediately
* [x] If the username and password fields aren't specified, instantiate the client but don't connect

Upon change of options:
* [ ] Compare the new options to the old to determine if anything really changed
* [ ] If anything changed, reconfigure the existing client with `ReconfigureOptionsAsync()`
* [ ] If `ReconfigureOptionsAsync()` returns `true` to indicate that a restart is required, set a flag somewhere and expose that on the UI

jpdillingham:
Closes #119

### DIFF
--- a/src/slskd/Application/Service.cs
+++ b/src/slskd/Application/Service.cs
@@ -22,6 +22,7 @@ namespace slskd
     using System.IO;
     using System.Linq;
     using System.Net;
+    using System.Text.Json;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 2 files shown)


jpdillingham:
Note that if the application is started without Soulseek credentials, it is currently impossible to connect without quitting the application, setting the credentials, and then restarting.  Ability to reconfigure options will be added in #97 and #122

PR #128, opened 2021-07-23
Rename 'File not found.' message to 'File not shared.'

mathiascode:
I believe the `File not found.` transfer status message is an oversight, since the Soulseek.NET library and official clients deal with `File not shared.` messages.

### DIFF
--- a/src/slskd/Application/Service.cs
+++ b/src/slskd/Application/Service.cs
@@ -413,7 +413,7 @@ namespace slskd
             if (!fileInfo.Exists)
             {
                 Console.WriteLine($"[UPLOAD REJECTED] File {localFilename} not found.");
-                throw new DownloadEnqueueException($"File not found.");
+                throw new DownloadEnqueueException($"File not shared.");
             }
 
             if (tracker.TryGet(TransferDirection.Upload, username, filename, out _))


jpdillingham:
Nice catch!

I had to review the history behind this; the code in Soulseek.NET is written so implementers can provide any message they want, but the 'File not shared.` message is meaningful: https://github.com/jpdillingham/Soulseek.NET/commit/571173c8ec7aa2b5097e09226484b4f218a33ac6

I know I played with this at one point and established that I could send whatever message I wanted and that it would be displayed in Qt, but I'm not sure how we're supposed to differentiate between queued transfers and rejected ones, since the rejected flag is set when queued as well.

Maybe the logic should be to treat anything other than "Queued." as rejected?

mathiascode:
> I'm not sure how we're supposed to differentiate between queued transfers and rejected ones, since the rejected flag is set when queued as well.
>
> Maybe the logic should be to treat anything other than "Queued." as rejected?

That was previously the case in Nicotine+, when we used the TransferRequest message for queuing downloads as well.

I recently ended up switching to the newer QueueUpload (peer code 43) system, which turned out to be simpler to implement in Nicotine+'s workflow at least. I believe it's been in use in official clients since the early 2000s.

jpdillingham:
Interesting!  I concluded at one point that `QueueUpload` was the older behavior but I wouldn't be surprised if I were wrong.  I'll take a look at switching it over, I think it would be fairly straightforward.

Regarding the different messages, I'm not sure.  I evidently saw something at one point that caused me to test for a partial match on "not shared" instead of the full string, so I probably noticed the variation.

mathiascode:
Regarding the "File not shared" message in Nicotine+, after checking some old code again, I believe it was just an oversight. Back in 2003, the dot variant was sent in response to QueueUpload messages, while the no dot-variant was sent in response to TransferRequest messages. Later on, someone removed the dot variant entirely, probably for better consistency in the GUI. In any case, I've reverted the message to `File not shared.`.

> Interesting! I concluded at one point that QueueUpload was the older behavior but I wouldn't be surprised if I were wrong. I'll take a look at switching it over, I think it would be fairly straightforward.

I'm not entirely when QueueUpload was first adapted. I know the message has existed for a long time, but these comments in Museek+ suggest NS 157 started using it:
https://github.com/eLvErDe/museek-plus/blob/d9defa2c6e69a22a108af33c6ab9c46b4ea28967/museekd/downloadmanager.cpp#L362
https://github.com/eLvErDe/museek-plus/blob/d9defa2c6e69a22a108af33c6ab9c46b4ea28967/museekd/peersocket.cpp#L233-L235

PR #139, opened 2021-08-04
Update README.md

ldexterldesign:
Both Soulseek & slskd `username` and `password` required for me to successfully login and search

If this is expected behaviour then both should be included in README

Hope this helps

Sincerely


jpdillingham:
Thanks, I totally overlooked this.  These values both default to `slskd`, but that's not documented either.  I definitely need to spend some time cleaning up the documentation.

ldexterldesign:
Yes, default values make most sense, just needs to be in the README I think

PR #141, opened 2021-08-06
Substitute share paths with a hash

### LINKED ISSUE #53: Obscure path to shared folder and ensure only backslashes are sent in browse and search results
See also: https://github.com/jpdillingham/Soulseek.NET/issues/492

Soulseek Qt uses only backslashes in paths, regardless of operating system, so slskd should too.

There's a bit of a privacy concern with exposing the full path to shared files; if a user is sharing something on a network drive, the path that other users see might look like:

```
\\SERVER\username\slskd-share\artist\album\track.dat
```

Assuming the root shared folder configured within slskd is `slskd-share`, paths sent to other users from Soulseek Qt would look like:

```
@@abcde\slskd-share\artist\album\track.dat
```

Rather than try and mimic this replacement (or hash, or whatever it is), the plan will be to just drop it and treat all paths as relative to the shared directory.  Down the road when support for multiple shared directories is added, each share will need to be given a unique name and that will become part of a "virtual" path used to determine which share a path is relative to.

jpdillingham:
This PR adds functionality to the shared file cache to mask all but the most immediate directory in shared directories.  

The path:

```/home/Foo/My Sensitive Filepath/Music/Shared```

Appears in browse and search results as:

```@@hash/Shared```

The hash is computed by first computing the SHA1 hash of the directory, then by converting the hash to base 64, removing numbers, converting all letters to lowercase, and then taking the first 5.

I'm pretty sure at this point that Soulseek (Qt at least) is expecting hashes that:

* Begin with `@@`
* Contain only lowercase letters (no numbers)
* Are exactly 7 characters in length (including `@@` and 5 additional characters)

Including numbers seems to only obscure the path up to the first number, and including uppercase characters seems to invalidate the hash entirely.  Including more than 5 characters results in characters 6 onwards being displayed.

Closes #53

### DIFF
--- a/src/slskd/Application/Application.cs
+++ b/src/slskd/Application/Application.cs
@@ -21,6 +21,7 @@ namespace slskd
     using System.Collections.Concurrent;
     using System.IO;
     using System.Linq;
+    //using System.Linq;
     using System.Net;
     using System.Reflection;
     using System.Threading;
@@ -592,7 +593,9 @@ namespace slskd
         private Task EnqueueDownloadAction(string username, IPEndPoint endpoint, string filename, ITransferTracker tracker)
         {
             _ = endpoint;
-            var localFilename = filename.ToLocalOSPath();
+            var localFilename = SharedFileCache.Resolve(filename).ToLocalOSPath();
+            Console.WriteLine($"Local: {localFilename}");
+
             var fileInfo = new FileInfo(localFilename);
 
             if (!fileInfo.Exists)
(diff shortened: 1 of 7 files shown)


jpdillingham:
It looks like Soulseek expects the hash to be letters only; any numbers make it appear in the UI.  I'll need a new hashing strategy that allows me to remove numbers while staying relatively unique

jpdillingham:
@mathiascode FYI, in case you ever need to know the hashing requirements.  I'll probably circle back on this at some point and allow users to specify virtual names for their shares, like Nicotine+ does.

Issue #173, opened 2021-08-30
Write logs to the application directory, instead of the executable directory


jpdillingham:
Logs should be written to the application directory, under `logs`, instead of next to the executable (as it is now).

Exouxas:
Just to clarify, the logs seem to be written to where the application is located under the folder `logs`. From looking at blame, this seems to have been correctly configured for a while?
![image](https://user-images.githubusercontent.com/26272249/135767901-662b09ae-587e-42b9-8a51-afceb2375d19.png)

jpdillingham:
Sorry the issue wasn't totally clear on this, by "application directory" I meant the directory configured [here](https://github.com/slskd/slskd/blob/master/src/slskd/Options.cs#L259),  This is where downloads, incomplete files and *.db files are stored, typically `%localappdata%\slskd` on Windows or `~/.local/share/slskd` on other platforms, unless overridden in options.

Should be a pretty small change [here](https://github.com/slskd/slskd/blob/master/src/slskd/Program.cs#L228)

Exouxas:
I'd like to try my hand at this if that's alright.

I've looked at the existing code, and if I understand it correctly there should be an instance of `Options` that can be used to get this "Application Directory". I've also found that the `Program` class has a usable instance called `OptionsAtStartup`, but the directories get verified after where the logging is initiated. Would this be an issue?

jpdillingham:
Closed by #209.  Thanks again!

PR #209, opened 2021-10-05
Changed path of logs from "exe path" to "application path"

Exouxas:
Made and tested change of path of logs. They are now placed under the application directory.

This is to very specifically fix issue #173, without changing when the directory is verified.

As written in #173: The `Program` class has a usable instance called `OptionsAtStartup`, but the directories get verified after where the logging is initiated. Should this be fixed in this same pull request?


Exouxas:
I have moved the code now, and made it use `Console.WriteLine`. I moved all 3 directory checks as to not fragment the logic (to keep readability up). Does this look like a good fix?

PR #245, opened 2021-11-09
Move the default configuration location to the App directory, create a Docker volume for /app

jpdillingham:
This PR refactors a number of small things about the configuration to set the stage for run-time editing of the YAML file from the application itself.  Chief among these changes is the move of the configuration file to the `app` directory (where the logs and database files are currently stored), and creation of a volume in the Docker container so that this data is all persisted.

Important to note is that this change _removes_ the ability to specify where the configuration file is stored.  The configuration file is now permanently fixed to the root of the `app` directory, which you can specify with:

* The `SLSKD_APP_DIR` environment variable
* The `--app` command line argument

The YAML file can no longer be used to specify the location.  If no value is supplied the directory defaults to the current location, which is `%localappdata%/slskd` on Windows and `~/.local/share/slskd` on Linux and macOS.

Functional tests:

* [x] App directory defaults to the expected location on Windows
* [x] App directory defaults to the expected location on Linux
* [x] App directory defaults to the expected location on macOS
* [x] Example config is copied to root of app directory on start if one does not exist
* [x] Config is loaded from `/app/slskd.yml`
* [x] Docker container creates volume and persists configuration


src/slskd/Options.cs
```diff
@@ -837,7 +833,7 @@ public class IntegrationOptions
             ///     Gets FTP options.
             /// </summary>
             [Validate]
-            public FTPOptions FTP { get; private set; } = new FTPOptions();
+            public FtpOptions Ftp { get; private set; } = new FtpOptions();
```
jpdillingham, line 836: The casing changed here because the underscore strategy for the yaml library in use would serialize this as `f_t_p`.  Pascal case is in alignment with .NET standard practices, also.

PR #259, opened 2021-11-13
Don't send empty picture in user info response

mathiascode:
Noticed this while viewing the user info of a slskd user. Since the picture is not null, the HasPicture parameter in the user info response message is set to true: https://github.com/jpdillingham/Soulseek.NET/blob/8f5e6582aee1dc17b473a11b5b9e056abdc9564e/src/UserInfo.cs#L36


jpdillingham:
Thanks! Someday I'll get around to actually implementing this :)

mathiascode:
No problem, take your time. There's always too much work involved in these clients. :)

PR #346, opened 2022-02-04
Add upload slot and bandwidth limiting for user groups

### LINKED ISSUE #127: Add ability to limit upload / download speed and slots
Another nice to have feature:
It would be nice to be able to limit speeds e.g. during working hours, as well as limit upload slots as in SoulSeekQT.

jpdillingham:
This PR adds slot and speed limits for both uploads and downloads, along with user groups and a proper upload queue.  The documentation for configuration has been updated, but I'll share the new section here in lieu of a duplicated explanation of the changes:

# Limits and User Groups

## Global

Global limits behave as a hard limit, additive across all groups.  These values should be set as high as practical for the environment in which the application is running; more granular controls should be defined at the group level.

A change to slot limits require an application restart to take effect, while speed limits can be adjusted at runtime.

|Command Line|Environment Variable|Description|
|----|-----|-----------|
|`--upload-slots`|`UPLOAD_SLOTS`|The limit for the total number of upload slots|
|`--upload-speed-limit`|`UPLOAD_SPEED_LIMIT`|The total upload speed limit|
|`--download-slots`|`DOWNLOAD_SLOTS`|The limit for the total number of download slots|
|`--download-speed-limit`|`DOWNLOAD_SPEED_LIMIT`|The total download speed limit|

#### **YAML**
```yaml
global:
  upload:
    slots: 20
    speed_limit: 1000
  download:
    slots: 500
    speed_limit: 1000
```

## Groups

User groups are used to control upload slots, speed limits and queue behavior on a per-user basis.

Each group has a priority, starting from 1, that determines the order in which groups are prioritized in the upload queue.  A lower number translates to higher priority.  Upload slots are granted to higher priority groups before lower priority groups.

Groups have a queue strategy, which can be either `FirstInFirstOut`, or `RoundRobin`.  This setting determines how uploads from multiple users in the same group are processed; `FirstInFirstOut` processes uploads in the order in which they were enqueued, while `RoundRobin` processes uploads in the order in which the user was ready to receive the upload.

Upload slots and speed limits configured at the group level can be used to create constraints in addition to global settings.  If group-level limits exceed global settings, global settings become the constraint.  Slots and speed limit settings default to `int.MaxValue`, effectively deferring to the global limits.

The general configuration for a group is as follows:

#### **YAML**
```yaml
upload:
  priority: 500 # 1 to int.MaxValue
  strategy: roundrobin # roundrobin or firstinfirstout
  slots: 10 # 1 to int.MaxValue
  speed_limit: 100 # in KiB/s, 1 to int.MaxValue
```

## Built-In Groups

The `default` built-in group contains all users that have not been explicitly added to a user defined group, are not priviledged, and that haven't been identified as leechers.

The `leechers` built-in group contains users that have been identified as leechers.

The `priviledged` built-in is used to prioritize users who have purchased priviledges on the Soulseek network.  This groups is not configurable, has a priority of 0 (the highest), a strategy of `FirstInFirstOut`, and can use any number of slots, up to the global limit.

It is not possible to explicitly assign users to built-in groups, but the priority, number of slots, speed and queue strategy can be adjusted (excluding `priviledged`).

#### **YAML**
```yaml
groups:
  default:
    upload:
      priority: 1
      strategy: roundrobin
      slots: 10
      speed_limit: 50000
  leechers:
    upload:
      priority: 99
      strategy: roundrobin
      slots: 1
      speed_limit: 100
```

## User Defined Groups

Any number of user defined groups can be added under the `user_defined` key in the group configuration.

User defined groups use the same configuration as built-in groups, but additionally allow a list of `members` containing the usernames of the users assigned to the group.

Users can be assigned to multiple groups, with their effective group being the highest priority (lowest numbered) group.  If a user has priviledges on the network, any explicit group membership is superceded, and their effective group is the built-in `priviledged` group.  If a user is explcitly assigned to a group, they will not be identified as a leecher.

#### **YAML***
```yaml
groups:
  user_defined:
    my_custom_group_name:
      upload:
        priority: 500
        strategy: roundrobin
        slots: 10
        speed_limit: 100
      members:
        - bob
        - alice
```

## Example

In the following example:

* All leechers share 1 slot among them, and can download at a maximum speed of 100 KiB/s.  Leechers can only download if there are fewer than 20 other uploads in progress to other users.
* Users that aren't leechers and that aren't in the `my_buddies` group (`default` users) share 10 slots among them and can download at the global maximum speed of 1000 KiB/s, but only if fewer than 20 upload slots are being used by users in the `my_buddies` group.
* Users that are in the `my_buddies` group share 20 slots among them and can download at the global maximum speed.  10 upload slots are reserved for users in this group, and users `alice` and `bob` are members.

```yaml
global:
  upload:
    slots: 20
    speed_limit: 1000
  download:
    slots: 500
    speed_limit: 1000
groups:
  default:
    upload:
      priority: 500
      strategy: roundrobin
      slots: 10
  leechers:
    upload:
      priority: 999
      strategy: roundrobin
      slots: 1
      speed_limit: 100
  user_defined:
    my_buddies:
      upload:
        priority: 250
        queue_strategy: firstinfirstout
        slots: 20
      members:
        - alice
        - bob
```

Closes #127


BachoSeven:
Thanks for adding this, it's really useful!

PR #392, opened 2022-03-19
Allow user description to be customized

Daniel15:
Title says it all. 😄 

## Test Plan
Tested by viewing the description in Nicotine+

Default:
image

Customized in YAML file:
image

Multiple lines in YAML file:
image


src/slskd/Options.cs
```diff
@@ -694,6 +694,15 @@ public class SoulseekOptions
             [RequiresReconnect]
             public string Password { get; init; } = null;
 
+            /// <summary>
+            ///     Gets the description of the Soulseek user.
+            /// </summary>
+            [Argument(default, "slsk-description")]
+            [EnvironmentVariable("SLSK_DESCRIPTION")]
+            [Description("user description for the Soulseek network")]
+            [RequiresReconnect]
```
jpdillingham, line 703: Since this is resolved dynamically and the `Options` is a shortcut for `OptionsMonitor.CurrentValue`, changes are reloaded on the fly and won't require a reconnect to update.

jpdillingham:
Thanks so much for this!  One minor tweak noted above.

src/slskd/Options.cs
Daniel15: Good catch! I'll update this soon.

jpdillingham:
I'm anxious to get this in, so I'll merge it as is and remove `RequiresReconnect` in a follow-up.  Thanks again!

PR #394, opened 2022-03-22
Add metadata to files when scanning shares

### LINKED ISSUE #136: Add metadata to shared files
The information collected about shared files is currently limited to the name and size only.  Select a third party library (or, if it is straightforward enough, write something from scratch) to inspect files to determine the bitrate, sample rate and length.

This needs to be as fast as possible, as adding this feature is likely to double or triple the amount of time it takes for the application to scan shares.

This should support at least as many formats as Soulseek NS/Qt support, including .mp3, .flac, .wav, and .m4a as top priorities.  Support for more esoteric formats can be added later, if it isn't straightforward to do so now.

jpdillingham:
Closes #136


src/slskd/Shares/SoulseekFileFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    Log.Warning("Failed to read metadata from file '{Filename}'; the file is an unsupported format or may be corrupt ({ExceptionType})", filename, ex.GetType().Name);
+                }
+                finally
+                {
+                    file?.Dispose();
+                }
+            }
+
+            return new File(code, maskedFilename, size, extension, attributeList);
```
mathiascode, line 94: Worth noting about the extension:
- SoulseekQt always sends an empty extension in file responses
- Soulseek NS only shows attributes if the extension value is "mp3"

https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/SLSKPROTOCOL.md#data-order-101

Also the various variations of the attribute list (I'll document them properly at some point):
https://github.com/nicotine-plus/nicotine-plus/blob/6af10d295dcad4c6365a3e2107b7279cf11f46c1/pynicotine/utils.py#L253-L376

jpdillingham: I'm doing some final testing this morning and have noticed that other clients are kind of all over the map in terms of what attributes they return and how they display incoming attributes.

I had already noticed that sample rate is discarded if bit depth is 0, so I only report either of those if bit depth is >0, but I'm not sure that's correct.

I hadn't anticipated that attributes might be positional so I may need to go back to Soulseek.NET and change the way results are parsed.

With the code the way it is now, Nicotine+ won't display any attributes for flac, probably because I'm not sending them in the expected order.

I'll have to keep digging and see if I can find a happy medium.  I guess I could just match the behavior of Qt but it isn't compatible with NS.

jpdillingham: Only NS and Nicotine+ seem to care about the VBR attribute.  Unfortunately it doesn't look like TagLib has an easy way to get this, either.

jpdillingham: Old discussion about attributes: https://www.slsknet.org/news/node/2945

I agree that sample rate isn't very useful for mp3s, so I'll continue to leave that out.  This also sheds some light on why Qt doesn't display VBR information.

mathiascode: The attribute list is a mess since SoulseekQt was introduced, in my opinion. There's also a less common case where SoulseekQt only sends a single attribute, which can be either the bitrate or length (possibly also sample rate or bit depth, don't remember), but I haven't investigated how SoulseekQt identifies these yet. It's obviously not position based in this case, but it's difficult to say exactly what SoulseekQt does.

jpdillingham: I reordered attributes and are adding them selectively to ensure compatibility with Nicotine+; my results are now consistent with other flac results in search.

For what it's worth, the behavior of Qt and NS doesn't change if the attributes are re-ordered, so they both seem to respect the attribute id instead of using positional logic.  The 'place in attributes' field is actually an enum: https://github.com/jpdillingham/Soulseek.NET/blob/master/src/FileAttributeType.cs

I can't remember how I figured this out but it holds true.

jpdillingham: I have never seen an attribute with value 3, otherwise my code would throw.  I'm not sure why I was so sure I'd never see one; I should probably refactor this to produce a warning.

mathiascode: I wish I had looked into this earlier... I always assumed it was position-based (and the type/ID values indicated the position) since a previous Nicotine+ developer had made this assumption.

I've changed the code in Nicotine+ to check the attribute type instead of position: https://github.com/nicotine-plus/nicotine-plus/commit/18c98d8e9e2525f67a01896eadb0ff395050e18b

mathiascode: I've documented attribute combinations used by different clients here: https://github.com/nicotine-plus/nicotine-plus/commit/4b3905c49f1ca797a2f3babdce918b83baab87ec

Feel free to add slskd when it's ready, or any other combinations that are missing.

mathiascode: @jpdillingham I tested sending different attributes to a Soulseek NS client, and figured out that attribute type 3 is reserved for encoder information. It doesn't seem to be used though, Soulseek NS only shows "Encoder:" in the UI if the file path doesn't end with `mp3`, and SoulseekQt doesn't implement this attribute at all.

https://github.com/nicotine-plus/nicotine-plus/commit/8d703150cde6ec50a326911cfd3155470b69ac6c

PR #423, opened 2022-04-19
Disambiguate transfer size when less than 1 MB has been transferred

### LINKED ISSUE #421: Invalid units in Size column of Downloads page
In the third row 722.22 seems like kilobytes.

image

jpdillingham:
Also changes `kbps` to `Kbps` to make size and bitrate labels consistent.

Closes #421


droserasprout:
Maybe decimal places should be fixed to some value to avoid "jumping" between indentations, e.g. `3.1 -> 3.15`?

jpdillingham:
@droserasprout thanks again for the suggestions!

PR #428, opened 2022-04-19
Make user cards in search and transfer lists foldable

droserasprout:
Not sure this feature is what we need right now - I mostly wanted to say hi 👋

I aim to dedicate a significant part of my vacation free time to this project and would be happy to co-maintain it in the long term.

Here's what's going on here:

* Added fold/unfold button to `TransferGroup`, ~~`TransferList`~~ and `Response` components.
* Removed close button from `Response` component.
* ~~Replaced `TransferList`'s `Header` with `Card.Header` just to see how it looks.~~

|||
|-|-|
| image | image |
| image | image |

Checklist:

* ~~Do we need to fold directories too, or just users?~~ Reverted for now.
* ~~Implement fold-all/unfold-all buttons~~
* ~~Is there any meaning in making directory headers smaller?~~
* Are there any ways to reduce copypaste here? My first React PR, lol.


jpdillingham:
Thanks!  This change was actually in the back of my mind in support of the [Guided Search](https://github.com/slskd/slskd/milestone/3) milestone, which would include folding responses like this and adding some indication of how well the response 'fit' the request.

I think adding the fold/unfold functionality in your screenshot is a good start, along with a setting that lets users choose whether things should be folded or unfolded by default (perhaps just a toggle along with search options would be a good start)

I haven't looked at the code yet but will this evening.

droserasprout:
image
image

jpdillingham:
Nothing jumps out as being overly copy-pasted so I'm guessing you cleaned whatever that was up today.

There is currently an `x` icon on the right to discard results and it looks like that's been removed, was that intentional?

I see some nitpicky stuff that I could comment on but I'll wait for you to let me know when you're ready :)

droserasprout:
I removed the close button intentionally as being superseded by folding. It should be enough within a single search. But maybe we should add an "Ignored Users" field, persistent between searches, to the top bar and the close button would trigger expanding that list?

jpdillingham:
I think the close button is still warranted; I use it sometimes to pick through results, rather than continually clicking the button to reveal more results.  Closing the top result shifts everything up.

src/web/src/components/Search/Response.js
```diff
@@ -55,11 +59,15 @@ class Response extends Component {
         });
     }
 
+    toggleFolded = () => {
+        this.setState({'isFolded': !this.state.isFolded});
```
jpdillingham, line 63:
```suggestion
        this.setState({ isFolded: !this.state.isFolded });
```

src/web/src/components/Search/Response.js
```diff
@@ -28,13 +28,17 @@ class Response extends Component {
...
+        isFolded: this.props.isInitiallyFolded,
     }
 
     componentDidUpdate = (prevProps) => {
         if (JSON.stringify(this.props.response) !== JSON.stringify(prevProps.response)) {
             this.setState({ tree: buildTree(this.props.response) });
         }
+        if (this.props.isInitiallyFolded !== prevProps.isInitiallyFolded) {
+            this.setState({ isFolded: this.props.isInitiallyFolded });
+        }
```
jpdillingham, line 41: this is the correct pattern when using class-based components and lifecycle functions.  functional components are the way moving forward; anything new should be functional, and older class-based components like this will eventually be rewritten.

i'm only mentioning this since you said you were new to react, and i thought it was worth clarifying.  couple of links, if you are interested:

https://reactjs.org/docs/components-and-props.html
https://www.pluralsight.com/guides/prop-changes-in-react-component

jpdillingham:
One fairly minor nit, but otherwise looks good.  I'm open to see how it feels without the close button, it can be added back later.

jpdillingham (approved):

jpdillingham:
I went ahead and pushed it along.  @droserasprout thanks again!

PR #446, opened 2022-04-23
Refactor bootstrapping to use .NET 6 conventions

### LINKED ISSUE #444: Refactor WebHost.CreateDefaultBuilder to use whatever .NET 6 wants you to use
It wouldn't be a major release of .NET if the team didn't _once again_ change how the web host is bootstrapped.

The code is functional, at least, but `UseSerilog()` has been marked as obsolete and is causing a warning.

I tried for about half an hour to figure out how all of the following functionality can be retained using `WebApplication.CreateBuilder()`, and it doesn't look like there is a straightforward path.

```c#
var webHost = WebHost.CreateDefaultBuilder(args)
    .SuppressStatusMessages(true)
    .ConfigureAppConfiguration((hostingContext, builder) =>
    {
        builder.Sources.Clear();
        builder.AddConfigurationProviders(EnvironmentVariablePrefix, ConfigurationFile);
    })
    .UseSerilog()
    .UseUrls()
    .UseKestrel(options =>
    {
        logger.Information($"Listening for HTTP requests at http://{IPAddress.Any}:{OptionsAtStartup.Web.Port}/");
        options.Listen(IPAddress.Any, OptionsAtStartup.Web.Port);

        logger.Information($"Listening for HTTPS requests at https://{IPAddress.Any}:{OptionsAtStartup.Web.Https.Port}/");
        options.Listen(IPAddress.Any, OptionsAtStartup.Web.Https.Port, listenOptions =>
        {
            var cert = OptionsAtStartup.Web.Https.Certificate;

            if (!string.IsNullOrEmpty(cert.Pfx))
            {
                logger.Information($"Using certificate from {cert.Pfx}");
                listenOptions.UseHttps(cert.Pfx, cert.Password);
            }
            else
            {
                logger.Information($"Using randomly generated self-signed certificate");
                listenOptions.UseHttps(X509.Generate(subject: AppName));
            }
        });
    })
    .UseStartup<Startup>()
    .Build();
```

jpdillingham:
Closes #444


src/slskd/Program.cs
```diff
@@ -154,6 +177,7 @@ public static class Program
 
         private static IConfigurationRoot Configuration { get; set; }
         private static OptionsAtStartup OptionsAtStartup { get; } = new OptionsAtStartup();
+        private static ILogger Log { get; set; } = new ConsoleWriteLineLogger();
```
jpdillingham, line 180: This is primarily an improvement for developers.  The logger isn't configured until later, but messages are sent to stdout early on.  By splitting the logger configuration out to `ConfigureGlobalLogger()` it is less clear that the logger shouldn't be used until after that is executed.

Adding this generic implementation as a default allows the logger to be used prior to configuration.

Also worth mentioning that a lot of this was also due to the fact that `Log` needed to become a module level property when the existing code was split out.

Issue #441, opened 2022-04-22
System menu tabs should have separate URLs


droserasprout:
Separate URLs will make UI behavior more consistent plus allow to surf history with the "Back" button

https://stackoverflow.com/questions/57780893/how-can-i-make-semantic-ui-react-tab-click-use-a-react-router-navlink-to-update

jpdillingham:
Agreed.

jpdillingham:
This was closed by #452

PR #450, opened 2022-05-03
Fix bug preventing routes with more than one segment from working

jpdillingham:
I have been working on searches and fighting with this for the last two days.  The conclusion is that the UI can't operate properly behind a reverse proxy (when `url_base` is set) if it doesn't explicitly know which part of the current url is `url_base`.  There may be a clever way to deduce this that I'm missing, but for now I've added an unauthenticated endpoint `GET /urlbase` that allows the UI to explicitly fetch this when the `App` component is loading.  This allows it to explicitly prefix routes properly.

As a test I've implemented #441 and it works locally, but we'll see once I get it deployed.

### DIFF
--- a/src/slskd/Core/API/Controllers/ApplicationController.cs
+++ b/src/slskd/Core/API/Controllers/ApplicationController.cs
@@ -15,6 +15,8 @@
 //     along with this program.  If not, see https://www.gnu.org/licenses/.
 // </copyright>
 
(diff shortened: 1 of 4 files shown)


jpdillingham:
It worked, for the most part.  The connection to the logs hub is broken though so I don't advise that anyone uses this version of the canary.

PR #467, opened 2022-05-17
Redesign Search views

### LINKED ISSUE #84: Add an interface to view and delete saved searches
#81 persisted search context and results to disk, but lacked a way to retrieve them after a new search had been started.

* [ ] Add a way to view a list of saved searches
* [ ] Add a way to view the results from a saved search
* [ ] Add a way to delete a saved search
* [ ] Add an option to set the maximum time to save searches, and some automation to automatically delete them when they have expired.

### LINKED ISSUE #198: Add the ability to cancel search requests that are in progress
It is easy to press enter and search for a typo, e.g. "beach boys pet souds". This would allow the user to cancel the search and not have to wait for it to complete. It doesn't seem there is any way to do this at the moment.

### LINKED ISSUE #246: Bug: empty search results in repeated searches happening
If you search "" then slskd gets stuck in a search loop.
Would be solved by https://github.com/slskd/slskd/issues/198.
I was able to fix this by deleting `search.db`.

### LINKED ISSUE #249: Add ability to have multiple concurrent searches
It would be nice to be able to open a new tab and load slskd and make another search e.g. when waiting for long searches.

### LINKED ISSUE #255: Refactor persistence of search responses
I've been profiling memory for the past 12 hours and have been randomly searching for terms for which I know large numbers of responses are returned.  When memory is constrained, several out of memory exceptions have been thrown while updating the search database when the search is complete, which would be when the results are saved to the database.

Think about the best way to save this data.

jpdillingham:
This PR allows multiple concurrent searches, and replaces the existing search display with a list:

list

Historical searches can be accessed by clicking the search term in the list:

detail

Searches can now be stopped while in progress by clicking the 'stop' icon to the right of the search in the list, or with the 'stop' button on the detail view.  Completed searches can be deleted by clicking the 'trash' icon in either location.

Closes #84 
Closes #198 
Closes #246
Closes #249
Closes #255


src/slskd/Common/RateLimiter.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace slskd
+{
+    using System;
+    using System.ComponentModel;
+
+    /// <summary>
+    ///     Ensures a minimum interval between successive invocations of a delegate.
+    /// </summary>
+    public class RateLimiter : IDisposable
```
jpdillingham, line 26: This was introduced to cap the frequency of SignalR updates

PR #517, opened 2022-06-17
Adding the ability to collapse each folder in a result.

mattgoldspink:
I started using slskd from my phone and the amount of scrolling I have to do is a lot, especially when I know the result is not the match I want based on the title of the folder. So I added the ability to toggle the state of the folders in both the transfer lists and the more generic FileList.

I had to convert FileList from a functional component into a class so it could manage it's own state.


jpdillingham:
Thanks for this one, too!

I've been trying to move away from class-based components to functional components that use hooks for state.  I'll add some suggestions inline to give you an idea how that would look.

src/web/src/components/Shared/FileList.js
```diff
@@ -10,68 +10,85 @@ import {
...
-              </Table.Row>
-            )}
-          </Table.Body>
-        </Table>
-      </List.Item>
-    </List>}
-  </div>
-);
+class FileList extends React.Component {
+  state = {
```
jpdillingham, line 14:
```suggestion
  const [folded, setFolded] = useState(false);
```

src/web/src/components/Shared/FileList.js
```diff
@@ -10,68 +10,85 @@ import {
...
+    isFolded: false,
+  };
+
+  toggleFolded = () => {
+    if (!this.props.locked) {
+      this.setState({'isFolded': !this.state.isFolded});
```
jpdillingham, line 20:
```suggestion
      setFolded(!folded)
```

jpdillingham:
Those suggestions didn't turn out as clear as I had imagined.  A better example is the `Info` component [here](https://github.com/slskd/slskd/blob/master/src/web/src/components/System/Info/index.js)

Happy to collaborate if you'd like!

src/web/src/components/Shared/FileList.js
mattgoldspink: It's been a while since I've touched React, so I've not had a chance to try `useState` yet. Looks like this is my opportunity! Let me take a stab to revert this one back to a functional component.

mattgoldspink:
Thanks for the feedback. That was easy enough to switch back to a functional component. The only unknown I had was whether to toggle for a locked group just because there's no obvious icon to use. I figured that given the default is to hide locked ones it's less of an issue, but happy to change that if you have any suggestions

jpdillingham:
Looks awesome, thanks!

PR #518, opened 2022-06-17
Tracks that last longer than an hour don't show that.

mattgoldspink:
I've fixed up the formatSeconds to test if it's longer than an hour then show the extra portion.

### DIFF
--- a/src/web/src/lib/util.js
+++ b/src/web/src/lib/util.js
@@ -3,6 +3,9 @@ export const formatSeconds = (seconds) =>
   if (isNaN(seconds)) return '';
   var date = new Date(1970,0,1);
   date.setSeconds(seconds);
+  if (seconds >= 3600) {
+    return date.toTimeString().replace(/.*(\d{2}:\d{2}:\d{2}).*/, '$1');
+  }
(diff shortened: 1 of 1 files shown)


mattgoldspink:
Screen shot of how it looks now:

image

jpdillingham:
Thanks, I didn't realize this was an issue.  I'll merge this as is but if it ends up causing formatting issues due to the increased width I may tweak it to display total minutes instead (e.g. `77:38` instead of `01:17:38`)

At some point I'm going to update tables so they scroll horizontally on small screens so column width will be less of an issue.

mattgoldspink:
> Thanks, I didn't realize this was an issue. I'll merge this as is but if it ends up causing formatting issues due to the increased width I may tweak it to display total minutes instead (e.g. `77:38` instead of `01:17:38`)

Thanks for the release with these changes in. I was thinking about what you said about showing it as just minutes to deal with the width, but I think when you hit `100:00` then you're not too far from an additional `01:40:00`. A simpler tweak might be to drop the leading zero and just have `1:40:00` which means you're just adding space for `:`. Happy to tweak my existing fix if you feel that would be better.

PR #519, opened 2022-06-17
Add 'minbitdepth' filter for search results

### LINKED ISSUE #514: Feature request: more detailed filter for lossless files
In the "lossless" category there are multiple options for a bit depth/bitrate a file can have. Most .flacs are 16/44. But some are 24/96 or 24/192. It would be nice to have an option to filter in search results for instance all files with 24 bit depth.

jpdillingham:
Used in conjunction with `islossless`, `minbitdepth:24` can be specified to limit search results to only those with at least 24 bit depth.

Closes #514


jpdillingham:
CI is failing on:

```
/home/runner/.dotnet/sdk/6.0.100/NuGet.targets(130,5): error : '0.16.4.65534-0093702' is not a valid version string. (Parameter 'value') [/home/runner/work/slskd/slskd/src/slskd/slskd.csproj]
```

Which happens if the SHA is all numeric.  This is a pretty rare issue and safe to ignore.

PR #528, opened 2022-06-24
Focus on username input field on load.

### LINKED ISSUE #507: Set focus on username input when the login page loads
Pretty straightforward; when I navigate to the app and am not logged in, my cursor should be placed in the username input.

mattgoldspink:
For some reason if I use Form.Input for the component the ref is always undefined, so had to switch it to Input.

Fixes #507


jpdillingham (approved):

jpdillingham:
Thanks!  I think semantic-ui-react is a bit inconsistent with `ref` handling.

PR #622, opened 2022-08-31
More accurately report the number of used and available upload slots

jpdillingham:
Thanks for the nudge @mathiascode

### DIFF
--- a/src/slskd/Application.cs
+++ b/src/slskd/Application.cs
@@ -1038,7 +1038,7 @@ namespace slskd
                 // wait until my first download starts.
                 var info = new UserInfo(
                     description: Options.Soulseek.Description,
-                    uploadSlots: group.Slots,
+                    uploadSlots: Math.Min(group.Slots, Transfers.Uploads.Queue.GlobalSlots),
                     queueLength: forecastedPosition,
                     hasFreeUploadSlot: forecastedPosition == 0);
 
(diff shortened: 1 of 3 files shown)


src/slskd/Application.cs
```diff
@@ -941,7 +943,7 @@ private async Task<SearchResponse> SearchResponseResolver(string username, int t
                         Client.Username,
                         token,
                         uploadSpeed: State.CurrentValue.User.Statistics.AverageSpeed,
-                        freeUploadSlots: forecastedPosition == 0 ? 1 : 0,
+                        freeUploadSlots: group.Slots - group.UsedSlots,
```
mathiascode, line 946: This doesn't look right. It's supposed to be a boolean value indicating if there are free slots.

jpdillingham: I'm wondering what I saw that lead me to believe this.  I'm reading a byte, but there's also some logic after that to conditionally read a 32 or 64 bit integer: https://github.com/jpdillingham/Soulseek.NET/blob/master/src/Messaging/Messages/Peer/SearchResponseFactory.cs#L46-L64

mathiascode: > there's also some logic after that to conditionally read a 32 or 64 bit integer: 

I looked this up some months ago, and came to the conclusion that the assumption about a single 64 bit integer was incorrect (the official clients only displayed a maximum value of the 32 bit integer limit for the queue length): https://github.com/nicotine-plus/nicotine-plus/commit/71a92c061971a89412da7004d3828e891f3be965

jpdillingham: Thanks for the heads up!

PR #626, opened 2022-09-02
PKGBUILDs and systemd service files

droserasprout:
## TBD

- [ ] Directory /.local/share/slskd does not exist, and could not be created
- [ ] Trigger sysusers hook
- [ ] Use prebuilt binaries?
- [ ] `-git` flavor

### DIFF
--- a/pkgbuild/.gitignore
+++ b/pkgbuild/.gitignore
@@ -1,0 +1,4 @@
+src
+pkg
+*.tar.gz
+*.tar.zst
--- a/pkgbuild/PKGBUILD
+++ b/pkgbuild/PKGBUILD
@@ -1,0 +1,34 @@
+pkgname=slskd
+pkgver=0.16.20
+pkgrel=1
+arch=("x86_64")
+url="https://github.com/slskd/slskd"
+license=("GPL3")
+options=("strip")
+depends=("dotnet-runtime")
+makedepends=("npm" "dotnet-sdk")
+source=("$url/archive/$pkgver.tar.gz")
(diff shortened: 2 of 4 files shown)


entrider:
Thank you for doing a good job, keep it up!

aopom:
I'd like to share the systemd file I'm using in production. You may not find every option useful, but it can be interesting:
```systemd.service
[Unit]
After=network.target
Description=A modern client-server application for the Soulseek file sharing network

[Service]
Type=simple
User=slskd
Group=slskd
StateDirectory=slskd
ExecStart=/path/to/slskd --app-dir /var/lib/slskd --config /path/to/slskd.yml
Restart=on-failure
EnvironmentFile=/var/secrets/slskd

# Security-related options. See man `systemd.exec`

# May break the service if it has to write outside of /var/lib/slskd
# Allow-list paths with `ReadWritePaths=`
ProtectSystem=strict

# May break the service if it has to look in /home
# Allow-list paths with `ReadOnlyPaths=`
ProtectHome=true

# The ones below shouldn't break anything
LockPersonality=true
NoNewPrivileges=true
PrivateDevices=true
PrivateMounts=true
PrivateTmp=true
PrivateUsers=true
ProtectClock=true
ProtectControlGroups=true
ProtectHostname=true
ProtectKernelLogs=true
ProtectKernelModules=true
ProtectKernelTunables=true
ProtectProc=invisible
RemoveIPC=true
RestrictNamespaces=true
RestrictSUIDSGID=true
```

YourSandwich:
Hello, how is the Plan with this merge?
Will this AUR package be maintained, by the pull requester? Or is it planned to make the CI do the work automatically?

Issue #610, opened 2022-08-22
"Exception of type 'System.OutOfMemoryException' was thrown when scanning exceptionally large shares


khalimerot:
Hi, 
Each time i'm trying to scan one of my shared folder i got this error after a few minutes (between 20 & 30% of the scan) 

```
22:02:21	ERR	An unhandled exception has occurred while executing the request.
22:02:21	WRN	Encountered error during scan of shared files: "Exception of type 'System.OutOfMemoryException' was thrown.
22:02:21	ERR	Failed to scan shares.
```

I already have split the scan in multiple part, but this folder have +/- 700k files.

My server is a 64GB RAM, SLSKD is running in a docker, on Unraid.
I don't think I have a RAM limitation on this docker, and the RAM of the server is not full.

Scanning the same folders with realies/soulseek docker worked well, even if it was very long.

Any idea on how i can solve the problem? 

Thank you!

jpdillingham:
Thanks for reporting this.  Does the application continue running after this?

If you [enable debugging] by setting `DEBUG=true` in the envars for the container you should get a more detailed error, but I'll push up an update that will improve this shortly and you can just update instead.

khalimerot:
Here is the log with the debug: 


```
17:52:27 | WRN | Encountered error during scan of shared files: "Exception of type 'System.OutOfMemoryException' was thrown.
17:52:27 | ERR | Failed to scan shares.
17:52:27 | INF | Scanned 0% of shared directories.  Found 548944 files so far.
17:39:48 | DBG | Failed  to read metadata from file '"/shared/Stone  Oakvalley/AMIGA/soamc/soamc_mp3/Trackers_Modland/Jamsam/mod.el4_T01_ORG.mp3"';  the file is an unsupported format or may be corrupt  ("CorruptFileException")
17:32:09 | DBG | Failed  to read metadata from file '"/shared/Stone  Oakvalley/AMIGA/soamc/soamc_mp3/Trackers_Games/Spider-Man_02/mod.gmour_T01_MIX.mp3"';  the file is an unsupported format or may be corrupt  ("CorruptFileException")
17:32:09 | DBG | Failed  to read metadata from file '"/shared/Stone  Oakvalley/AMIGA/soamc/soamc_mp3/Trackers_Games/Spider-Man_02/mod.action_T01_MIX.mp3"';  the file is an unsupported format or may be corrupt  ("CorruptFileException")
[...]

PR #637, opened 2022-09-09
Refactor shared file cache to store all data on disk

jpdillingham:
Currently shared files are stored in memory within an SQLite database (for full text search) and in a dictionary (for metadata).  This is fast, but uses a lot of memory and large shares cause the application to crash, as reported in #610.

This PR refactors the cache to store all data on disk, and to load the cache from disk on start, negating the need to scan on every start.

These changes allow very large collections to be shared, but negatively impacts the performance of search and browse response creation.

A future update will add an option to use on-disk or in-memory storage so users can select the strategy that works best for their collection.

Remaining tasks:

* [ ] Review the sort order of various responses.  Files and directories should be returned to callers in alphabetical order
* [ ] ~~Add an option to allow the share database to be stored on disk (`/data/shares.db`) or in memory (`file:shares?mode=memory`)~~ Follow-up
* [ ] ~~When using an in memory database, back up to disk (`/data/shares.db`) when the scan is complete, and load from disk when the application starts~~ Follow-up

### DIFF
--- a/src/slskd/Application.cs
+++ b/src/slskd/Application.cs
@@ -369,22 +369,23 @@ namespace slskd
                 _ = CheckVersionAsync();
             }
 
-            //if (OptionsAtStartup.Flags.NoShareScan)
-            //{
-            //    Log.Warning("Not scanning shares; 'no-share-scan' option is enabled.  Search and browse results will remain disabled until a manual scan is completed.");
-            //}
-            //else
-            //{
-            //    _ = Shares.StartScanAsync();
-            //}
+            if (await Shares.TryLoadFromDiskAsync())
+            {
+                Log.Information("Loaded share cache from disk");
+            }
+            else
+            {
+                Log.Warning("Failed to load share cache from disk. A share scan is required.");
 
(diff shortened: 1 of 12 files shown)


jpdillingham:
There are a few open questions still namely:

* Does using `yield` when fetching collections perform better than returning a large data object? There's probably a tradeoff between speed and memory usage
* Is it better to fetch rows and sort in memory, or to include an `ORDER BY` in SQLite queries? 
* What's the best way to incorporate a regular table with one indexed with FTS5? Right now there's a join.  Would an `IN` subquery work better? Would it be better to put all of the columns in the FTS5 table and mark them `UNINDEXED`?

One of the columns in the metadata table stores the attribute list of each file as json, which must be deserialized when composing a response containing files.  There's an optimization opportunity that would have the file metadata stored as binary, in the format that would be sent directly over the network, similar to what is done in this PR with the browse response.

I don't think this level of optimization is necessary just yet, but I'm curious how the performance scales with very large (500k+) shares.

PR #661, opened 2022-09-25
Default scan workers to Environment.ProcessorCount

### LINKED ISSUE #659: Decrease the default share scanning worker count
This is currently 16, and it's causing issues with responsiveness of the application while the scan is running.  It might be best to detect the number of cores available on the host machine and use that as the default.

jpdillingham:
The previous default (16) was causing the application to become unresponsive during a scan, depending on the system the app was running on.  This PR uses what it believes is the processor count of the system, which should be the best for most people.

I guess I should note that the performance of the scan has a lot more to do with I/O performance than CPU so this might not make a lot of sense for some systems, but it's better than using a static value.

Closes #659

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -170,7 +170,7 @@ Scanning shares is an I/O intensive operation; all of the files and directories 
 are used to allow metadata to be read from several files concurrently. The scan generally gets faster as additional workers are added, but each worker also adds additional I/O pressure and increases CPU and memory usage. At some number of workers
 performance will start to get worse as more are added.  The optimal number of workers will vary from system to system, so if scan performance is important to you it will be a good idea to experiment to see what the optimal number is for your system.
 
-The default number of workers is 16.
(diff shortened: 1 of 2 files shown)


mathiascode:
What kind of files have you tested with? I wonder to what extent multiple workers help if you mostly share larger files.

https://jrtechs.net/programming/multi-threaded-file-io

jpdillingham:
I'm leaning towards this not being much of a help for most people, but I don't have any numbers to back that up.

When I was developing the feature I was using 16 threads and saw speed gains of 4-5x.  What I failed to consider was that I was scanning a network share, and therefore nearly all of the I/O was network bound.  Additionally my NAS is set up to stripe files across ~7 hard drives, and that spread actual disk I/O out quite a bit more than the average person would see.

Even with the thread count equal to the number of processors on the system, people are complaining about the UI becoming unresponsive.  I may decrease the default further to half the number of processors.

jpdillingham:
I guess I didn't actually answer your question! I tested using my music library which is almost all mp3 at v0 or 320.

PR #664, opened 2022-09-30
Fixed padding-top to 14px for transfers-header-segment

skar07:
Reduced the 6px of ```padding-top``` for ```transfers-header-segment```.
Added ```margin-top``` property to ```transfers-segment-icon``` class to position icons appropriately.


jpdillingham:
Thanks!

PR #667, opened 2022-10-01
Changes <Response> class component to functional component

skar07:
Refactored ```Response.js``` to convert ```<Response>``` to functional component.
Updated import of ```<Response>``` in ```SearchDetail.js```.

### DIFF
--- a/src/web/src/components/Search/Detail/SearchDetail.js
+++ b/src/web/src/components/Search/Detail/SearchDetail.js
@@ -11,7 +11,7 @@ import {
 import { sleep } from '../../../lib/util';
 import Switch from '../../Shared/Switch';
 import ErrorSegment from '../../Shared/ErrorSegment';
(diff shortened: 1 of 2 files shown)


jpdillingham:
Thanks!  Unfortunately the browser crashes when I make any change on the page.  I think the state is pretty tricky in this component and I'm remembering why I didn't just convert it instead of creating an issue.

skar07:
> Thanks! Unfortunately the browser crashes when I make any change on the page. I think the state is pretty tricky in this component and I'm remembering why I didn't just convert it instead of creating an issue.

Ah, I'll have a look into how I am handling the state in the refactor and see if I can find a workaround/fix for the crash.

jpdillingham:
I'd hate to have you spend a bunch of time on this; feel free to convert any other component instead, I think there's a ton of them that should be very straightforward.

PR #670, opened 2022-10-01
Display the correct buttons in the header when viewing a search that returned no results

### LINKED ISSUE #648: Viewing a search that received no results causes the 'delete' button to read 'stop'
Pressing the 'stop' button deletes the search.

A search with no results should have the 'delete' and 'search again' buttons.

vanogrid:
This PR should fix an issue with displaying a 'Stop' button in the SearchDetailHeader instead of 'Delete' and 'Search Again' when viewing a search that returned no results.

Closes #648


jpdillingham:
Thanks!

PR #679, opened 2022-10-05
Change version delimiter for hash to plus

### LINKED ISSUE #642: Builds fail if the short commit SHA of the head is only numbers
```
/home/runner/.dotnet/sdk/6.0.100/NuGet.targets(130,5): error : '0.16.21.65534-0146571' is not a valid version string. (Parameter 'value') [/home/runner/work/slskd/slskd/src/slskd/slskd.csproj]
```

This is a shortcoming of .NET tooling; it doesn't want to see a numeric suffix here.  I believe the rationale is that they expect this to be something like `-beta` or `-rc1`, so sticking the SHA here is a little unorthodox (though it shouldn't be).

When this happens I have to push an empty commit to either the feature branch or master to get it resolved.  It's pretty rare, but also pretty annoying.

Exouxas:
Should fix #642

Now conforms with SemVer 2 and shouldn't give errors if the short hash is only numbers

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -42,7 +42,7 @@ done
 if [ -z "$version" ]; then 
   tag=$(git describe --tags --abbrev=0)
   sha=$(git rev-parse --short HEAD)
-  version=$tag-$sha
+  version=$tag+$sha
 fi
 
 if [ "$dotnet_only" == true ]; then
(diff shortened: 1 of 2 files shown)


jpdillingham:
Thanks!

PR #680, opened 2022-10-05
Mutex implementation to check single instance

skar07:
### Description
- Implementation of mutex on running instance of slskd
- Check for pre-existing mutex and accordingly disallow multiple instance creation


jpdillingham:
Thanks!

I'm actually working on a feature that will require that I disable this in certain scenarios 😅 

I'll merge this as is and tweak it later though.

Issue #407, opened 2022-03-30
Append a numbered suffix to newly downloaded files if a file of the same name already exists


jpdillingham:
When a download is complete, check the destination folder for a file of the same name, and if present, append a numbered suffix to the file being moved.  If a suffixed file exists, increment the number.

For example:

* File.mp3
* File (1).mp3
* File (2).mp3

Add an option to allow users to opt-out of this behavior and overwrite existing files instead.

Related to #401

jpdillingham:
You've got the right idea; I think I'd put all of the relevant logic [here](https://github.com/slskd/slskd/blob/master/src/slskd/Transfers/Downloads/DownloadService.cs#L484-L502).  

You're right to question the performance of a filename prefix search; for large directories this would be pretty slow, and because downloads inside of slskd are concurrent, we'd run the risk of collisions anyway.

A better approach would be to check if `File.mp3` exists, and if so, save the new file as `File_638007399693622889.mp3`, where the number comes from `DateTime.UtcNow.Ticks`.  This is uglier than `(1)` but serves the same purpose and we don't have to count the existing files.

Let's not worry about FTP uploads; there's a setting that lets users choose to overwrite or skip and I think that's good enough.  Checking whether the file exists on a remote server would likely be too slow and flaky.

Exouxas:
I've added the duplicate filename logic here 1012360d2f27aea950eec055a74b62710adab029

I'm not certain how to add the option to use the default logic, would this be the correct way?
```cs
// in Program.cs
[Argument('o', "overwrite", "overwrite existing files")]
public static bool Overwrite { get; set; }
```
```cs
// in DownloadService.cs
while (File.Exists(destinationFilename) && !Program.Overwrite)
```

jpdillingham:
There's not a good spot for this option in the YAML yet, which usually dictates how options are laid out in general.  Let's make this non-configurable for now.

jpdillingham:
Closed by #682

We can revisit the opt-out later if someone asks.

PR #682, opened 2022-10-08
Add logic for duplicate filenames

Exouxas:
Related to #407 

Does not include opt-out option.


Exouxas:
I'm still learning how to do changes in previous git commits, tried to make it more "collected" but ended up with an extra commit...

Still though, decided to go with jpdillingham's suggestion regarding filenames. Should now handle filenames that contain periods, and added an extra if-statement that will slightly lower the processing time of non-conflicts and slightly raise the processing time of conflicts.

jpdillingham (approved):
Looks great to me! Thanks!!

Issue #633, opened 2022-09-05
"The ContentPath field specifies a non-existent directory '/home/user/wwwroot'"


jpdillingham:
This happens when starting the app from a directory other than the one containing the executable.  The code is:

```c#
Path.GetFullPath(OptionsAtStartup.Web.ContentPath);
```

This resolves `wwwroot` as a directory relative to the current working directory, when it should be relative to the application directory, using `AppContext.BaseDirectory`.

skar07:
Could I be assigned to this? Would love to work on it

jpdillingham:
Sure thing!

Originally the intent was to allow users to specify a path outside of the application directory, and now that I'm thinking about it this is a very bad idea since it could cause sensitive files on the host to be exposed if someone were to set it to something like `/`.

Instead we should just assume that the user has provided a path relative to the directory containing the application executable.  I've created #684 to make this intent more clear to users, and to prevent them from specifying absolute paths, but this change can be made independently.

For this one I think you can substitute

```c#
var contentPath = Path.GetFullPath(OptionsAtStartup.Web.ContentPath);
```

With

```c#
var contentPath = Path.Combine(AppContext.BaseDirectory, OptionsAtStartup.Web.ContentPath);
```

skar07:
Just wanted a clarification. Since with this change we'd be inherently taking a path relative to the directory, would it make #684 redundant, since we wouldn't need to add ```DirectoryIsRelative``` attribute to the ContentPath option in order to make the path relative?

jpdillingham:
Ideally we'd do both; make it programmatically impossible to use an absolute path (as you've done in #687!), and make it impossible to specify an absolute path in the options.

If we don't add `DirectoryIsRelative`, users can still specify absolute paths, but they'll get an error later, or maybe the web interface won't work and they might not be able to tell why.

jpdillingham:
Closed by #687.  Thanks!

PR #687, opened 2022-10-10
Update contentPath to use AppContext.BaseDirectory

skar07:
### Description
- This PR addresses #633 
- It makes changes to ```Program.cs``` and ```contentPath```
- It avoids using the absolute path which can allow for security issues, and instead lets the user give relative paths with respect to the application directory.


jpdillingham (approved):
Thanks!!

Issue #684, opened 2022-10-09
Add a `DirectoryIsRelative` validation attribute, similar to `DirectoryExists` and add the attribute to the ContentPath option


jpdillingham:
This should be as simple as checking [Path.IsPathRooted](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.ispathrooted?view=net-6.0) and failing validation if the result is `true`.

Once the attribute is added, add it to `ContentPath` in `Options`.

pn04:
If this has not been picked up by anyone, can I work on it?

jpdillingham:
In that case the default will be what's validated, so not necessary.  This is happening [here](https://github.com/slskd/slskd/blob/master/src/slskd/Program.cs#L333-L358); essentially the `Options` instance is materialized from all sources, bound to `OptionsAtStartup`, and then validated.

jpdillingham:
Closed by #694.  Thanks!

PR #694, opened 2022-10-12
added a DirectoryIsRelative validation attribute

pn04:
- This PR addresses #684 
- added a new `DirectoryIsRelative` attribute which checks if path is relative, and fails validation when it's not relative


jpdillingham (approved):
Awesome, thanks for including tests!

jpdillingham:
Aaaand here's where it gets tricky because the static `Path` methods have different behavior depending on the OS where they are executed.

I'm ok with deleting the tests.  If you really want to include them you'll have to add conditionals that use [IsOSPlatform()](https://stackoverflow.com/questions/38790802/determine-operating-system-in-net-core) (Windows and non-windows should be sufficient)

pn04:
I deleted the tests for now, could not figure out a clean way to run same test on any platform. Then thought about writing different tests for each platform, but I think xUnit 2.x does not support dynamically skipping tests.

PR #723, opened 2022-11-01
Normalize all paths shared with the Soulseek network to use backslashes

jpdillingham:
I thought I was in a good place with this, but now am encountering issues with the new network mode where the agent and controller are using different operating systems.  The path localization logic doesn't work for these because the masked paths saved in the shared file database use localized slashes.

This PR updates the scanner logic to save masked files with backslashes, regardless of operating system.  There's no longer a need to normalize paths on the way out to the network, so that logic is removed.

I'm really not sure why I didn't just do this in the first place; it's the much more logical option.  I'll probably find out when new bugs surface after this is merged 🙃

### DIFF
--- a/src/slskd/Shares/ShareScanner.cs
+++ b/src/slskd/Shares/ShareScanner.cs
@@ -232,7 +232,7 @@ namespace slskd.Shares
 
                             var share = Shares.First(share => directory.StartsWith(share.LocalPath));
 
-                            Repository.InsertDirectory(directory.ReplaceFirst(share.LocalPath, share.RemotePath), timestamp);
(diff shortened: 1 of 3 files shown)


mathiascode:
Makes sense, this is what we've been doing in Nicotine+ since 2016.

> There's no longer a need to normalize paths on the way out to the network, so that logic is removed.

What about existing databases of slskd users?

jpdillingham:
Good call; I'm thinking about how to handle this now.  Migrating existing databases/data has been a looming cloud I have been able to avoid so far but this will break everyone so I need a solution prior to the next release.

This was a lot easier when I was just throwing the database away at start :)

PR #854, opened 2023-01-19
Always show file size when downloads are cancelled or pending

### LINKED ISSUE #600: File size isn't shown for queued transfers
It might just be downloads.  The size is in the API response.  It displays as expected once the transfer starts.

mattgoldspink:
Instead of only showing the file size when some bytes have been downloaded always show it. That way a user knows the size of pending downloads too.

Fixes #600


jpdillingham (approved):
Thanks!

PR #881, opened 2023-02-19
Trying to speed up the performance of the Browse tab.

### LINKED ISSUE #317: Can't browse users with large libraries
When attempting to browse users with large libraries I am getting lots of errors.

First up while the browsing progress counts towards 100% there will be multiple 404 not found errors logged in the browser console like so:
```
GET http://192.168.1.3:55000/api/v0/users/username/browse/status 404 (Not Found) (anonymous) @ xhr.js:187
Uncaught (in promise) Error: Request failed with status code 404 @ createError.js:16
```

When it hits 100% the list of directories will be briefly shown before it almost instantly disappears and the following is logged in the console.
```
DOMException: Failed to execute 'setItem' on 'Storage': Setting the value of 'soulseek-example-browse-state' exceeded the quota. @ Browse.js:92
```
as a result of attempting this:
`localStorage.setItem('soulseek-example-browse-state', JSON.stringify(this.state));`

The default maximum localStorage size on Chrome and Firefox is 5mb (5120kb). For one user I was browsing, that JSON.stringify call generated over 11.8mb in a string. Perhaps you could maybe [use a compression library](https://pieroxy.net/blog/pages/lz-string/index.html) if it is necessary to store the data. That 11.8mb becomes 1.7mb using gzip on my box (though not using the linked lib). On the other hand the upper limit for a user's library size is potentially infinite so you're going to exceed this number sooner or later no matter what.

Let me know if you need the name of a user with a huge library to test on.

mattgoldspink:
It looks like there's a couple of issues with the current implementation which shows in Performance tab in Chrome:

1) getChildDirectories and in particular the split function. When you look at the code we're doing 3 filters on a reduced array each time, but always starting with the full list of directories. So for every child directory we iterate once over the entire array of all directories to remove 1 item, then for n-1 items we split them, then we find the ones who start with the parent path name. This is very inefficient.

To fix this I added one first iteration that will split the names on them all once and use that depth as a key into a Map and store an Array of the items at that depth. We could possibly go one step further and iterate the array store the d.name split on the last index of the seperator as a key to speed this up? But the new version is already orders of magnitude faster

2) The compression on saveState into localStorage is slow and blocking the UI thread. Given that the saveState doesn't affect the actual state in the component or whats rendered I've moved it out of the current tick of rendering and shifted it in most browsers to be done when the screen is idle (i.e. user is not doing anything). Or in the case of Safari falling back to the next tick.

This dropped render time for on one user with 6000+ directories from 1 minute to around 15s, of which a large chunk is the server calls.

Fixes #317

### DIFF
--- a/src/web/src/components/Browse/Browse.js
+++ b/src/web/src/components/Browse/Browse.js
@@ -100,10 +100,19 @@ class Browse extends Component {
     this.inputtext.inputRef.current.value = this.state.username;
     this.inputtext.inputRef.current.disabled = this.state.browseState !== 'idle';
 
-    try {
-      localStorage.setItem('soulseek-example-browse-state', lzString.compress(JSON.stringify(this.state)));
-    } catch (error) {
-      console.error(error);
+    const storeToLocalStorage = () => {
+      try {
+        localStorage.setItem('soulseek-example-browse-state', lzString.compress(JSON.stringify(this.state)));
+      } catch (error) {
+        console.error(error);
+      }
+    };
+    // Shifting the compression and safe out of the current render loop to speed up responsiveness
+    // requestIdleCallback is not supported in Safari hence we push to next tick using Promise.resolve
+    if (window.requestIdleCallback) {
+      window.requestIdleCallback(storeToLocalStorage);
+    } else {
+      Promise.resolve().then(storeToLocalStorage);
     }
   };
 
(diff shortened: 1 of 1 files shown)


jpdillingham:
Looks promising, thanks!  Once the release I just created is out I'll merge this and we can see what happens.

Storing the result in localstorage is always going to determine the ceiling for the size of response that can be handled; I need to refactor this to save it server side instead.  But that can come later.

Issue #898, opened 2023-04-14
Unable to remove requested search


technerdist:
Hi,

I accidentally searched for the letter "V", and received the following error:
![image](https://user-images.githubusercontent.com/83302492/232106613-f8eb03d9-8a78-42c3-b386-6cedaccdbd0f.png)

These searches are unable to be stopped, they are stuck in an endless loop. Cancelling does nothing, no way to remove the search. I managed to click search again many times, which subsequently created many, many more entries. How can these be removed without clearing all my app data?

jpdillingham:
Thanks for reporting this.  I'll need to revisit the underlying implementation in Soulseek.NET but this is almost certainly throwing an exception under the hood and not being properly reported.  

On the slskd side this is a visual bug only.

PR #911, opened 2023-05-26
fix: unable to remove completed searches

kaybi-gh:
This PR aims to fix issue #898.

Changes:
A search is now considered completed if search state is either one of these:
- Completed
- Cancelled
- TimedOut
- ResponseLimitReached
- FileLimitReached
- Errored

Added a try-catch around Client.SearchAsync to disallow bad requests to be eternally at search state 'Requested'.

~~We may need to find a way to clean-off requests that are already stuck at state 'Requested'.
Maybe we could periodically put searches that are at state 'Requested' with a 'DateTime.Now' - 'StartedAt' exceeding a configured timeout value?~~

Edit: I added a 'ForceCancel' on Searches that allows to cancel searches that are at state 'Requested' with a 'DateTime.Now' - 'StartedAt' exceeding the configured Soulseek.Connection.Timeout.Inactivity option when TryCancel returns false.
Meaning that users that have stucks requests will be able to cancel them then remove them manually.

### DIFF
--- a/src/slskd/Search/SearchService.cs
+++ b/src/slskd/Search/SearchService.cs
@@ -170,13 +170,6 @@ namespace slskd.Search
 
             List<SearchResponse> responses = new();
 
-            void UpdateAndSaveChanges(Search search)
-            {
-                using var context = ContextFactory.CreateDbContext();
-                context.Update(search);
-                context.SaveChanges();
-            }
-
             options ??= new SearchOptions();
             options = options.WithActions(
                 stateChanged: (args) =>
@@ -267,6 +260,31 @@ namespace slskd.Search
             }
 
             return false;
+        }
+
+        /// <summary>
+        ///     Forces a cancel on the specified search.
+        /// </summary>
+        /// <param name="search">The search to force cancel.</param>
+        /// <returns>The operation context.</returns>
+        public async Task ForceCancel(Search search)
+        {
+            if (search == default)
+            {
+                throw new ArgumentNullException(nameof(search));
+            }
+
+            search.EndedAt = DateTime.UtcNow;
+            search.State = SearchStates.Cancelled;
+            UpdateAndSaveChanges(search);
+            await SearchHub.BroadcastUpdateAsync(search);
+        }
+
+        private void UpdateAndSaveChanges(Search search)
+        {
+            using var context = ContextFactory.CreateDbContext();
+            context.Update(search);
+            context.SaveChanges();
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/web/src/components/Search/List/SearchActionIcon.js
```diff
@@ -7,7 +7,12 @@ const SearchActionIcon = ({ search, loading, onRemove, onStop,...props }) => {
     return (<Icon name='spinner' loading {...props}/>);
   }
 
-  if (search.state.includes('Completed')) {
+  if (search.state.includes('Completed')
```
jpdillingham, line 10: the state flags end up being serialized as a comma-separated list, so this value will be something like `Completed, Cancelled`.  the state is guaranteed to contain the `Completed` flag when it's done, so we shouldn't need to include the different sub-states here

src/slskd/Search/Types/Search.cs
```diff
@@ -34,7 +34,13 @@ public class Search
         public Guid Id { get; init; }
 
         [NotMapped]
-        public bool IsComplete => State.HasFlag(SearchStates.Completed);
+        public bool IsComplete =>
```
jpdillingham, line 37: same as below; searches (and transfers 🙃) are guaranteed to get the `Completed` flag regardless of how it ends, so checking for `Completed` is enough

src/slskd/Search/SearchService.cs
```diff
@@ -198,13 +191,25 @@ void UpdateAndSaveChanges(Search search)
                     UpdateAndSaveChanges(search);
                 }));
 
-            var soulseekSearchTask = Client.SearchAsync(
+            Task<Soulseek.Search> soulseekSearchTask;
+            try
+            {
+                soulseekSearchTask = Client.SearchAsync(
```
jpdillingham, line 197: this is a subtle bug that i should have been more aware of.  as you pointed out in discord, the `SearchAsync` method is actually not async under the hood, it only returns a `Task`.  most of the public methods in Soulseek.NET are like this to disambiguate the behavior of guard clauses, which are designed to throw immediately at the call site, instead of when the resulting task is awaited.

we'll want to keep an eye out for similar cases.  nice catch!

src/slskd/Search/SearchService.cs
```diff
@@ -256,5 +261,30 @@ public bool TryCancel(Guid id)
 
             return false;
         }
+
+        /// <summary>
+        ///     Forces a cancel on the specified search.
+        /// </summary>
+        /// <param name="search">The search to force cancel.</param>
+        /// <returns>The operation context.</returns>
+        public async Task ForceCancel(Search search)
```
jpdillingham, line 270: after reviewing the existing search code i agree that this is necessary right now, but long term i'll create an issue to refactor searches so it isn't possible for them to fail and remain "stuck"

PR #961, opened 2023-09-06
fix: use alpine 3.18 and debian bookworm in docker build

onedr0p:
Updating some dependencies here, and thank you for this amazing app. I've been using soulseek on my mac with nicotine for years but this makes it so I can just run it on my server instead <3


jpdillingham:
Thanks for this!

PR #964, opened 2023-09-13
Fixes docker.md yml shares

Natooz:
👋

This PR just fixes the yaml examples in the docker documentation for the shares.
I came through this while setting up the image.


jpdillingham:
Thanks, nice catch!

Issue #505, opened 2022-06-09
Transfer size column is too narrow


jpdillingham:
![image](https://user-images.githubusercontent.com/17145758/172871984-543af64d-1a5f-496d-90b0-b39975112a74.png)

I thought I tested this with some worst-case-scenario values, but I guess not.  I believe something like `1000.00 Bytes/1000.00 Bytes` would be the longest possible value here.

This is taking up quite a bit of real estate on the screen and I'm questioning whether it is worthwhile to display both the bytes downloaded and size, especially when the objectively more interesting transfer speed isn't present at all.

I'm not sure what to do here exactly, but these values shouldn't be able to wrap.

ckbaudio:
You could truncate bytes and kilobytes by removing one or both decimals.

skar07:
Would it be fine if I work on this?

jpdillingham:
@skar07 sure thing!

I think the ideal solution is to make sure the suffix only appears once, so `1.38 KB/1.38 KB` would appear as `1.38/1.38 KB`.  I'm not sure why some sizes do this and `KB` and `Bytes` don't.  It's probably something pretty simple.

paverett:
Hi I'm looking to contribute and think I've got a solution, let me know what you think.

`Bytes` should be removed and replaced with the short-form `B`. Then we should only use one decimal place for both the size and transfer numbers. Finally the transfer number should always match whatever unit is displayed for the size so that it doesn't have to display two units at any time. example: (100 KB/100 MB) would be (0.1/100MB)

If you think this will work I can get a PR out shortly.

jpdillingham:
@paverett sounds great! I'll assign it to you.

PR #974, opened 2023-09-27
fix: transfer size column is too narrow

paverett:
This PR aims to resolve issue #505.

- Transfer and Size numbers reduced to 1 decimal.
- Bytes replaced with B
- Transfer number always matches the unit of the size number.


jpdillingham (approved):
Thanks!

paverett:
@jpdillingham Apologies I didn't run npm lint. I ran it with lint:fix and pushed the changes up. Should fix the build.

jpdillingham:
No problem at all, thanks again!

PR #1000, opened 2023-11-13
Minor fix in the user-defined queue strategy parameter

FaGuoMa:
Minor typo in the default slskd config (`queue_strategy` vs `strategy`


jpdillingham:
Thanks!

Issue #473, opened 2022-05-20
update docker docs to include the --user and user: flags


davidzx:
after an update i had to figure out again how to use the user flags so files are owned by the user and not root. a snippet in the docker docs would help

something like
For the files to be owned by the user and not root you can set the user flag

docker run
user pgid:uuid(1000:1000)
```
docker run -d \
  -p 5000:5000 \
  -p 5001:5001 \
  -p 50000:50000 \
  -v /var/slskd:/app \
  -v /home/JohnDoe/Music:/music \
  -v /home/JohnDoe/eBooks:/ebooks \
  -e "SLSKD_SHARED_DIR=/music;/ebooks" \
  --name slskd \
  --user 1000:1000
  slskd/slskd:latest
```

or yml
user: 1000:1000
```
version: "2"
services:
  slskd:
    image: slskd/slskd
    container_name: slskd
    user: 1000:1000
```

jpdillingham:
Thanks for raising this.  It's been discussed a few times in issues but the docs have never been updated to reflect the outcome of these conversations.

emes81:
Could the docker-compose on [README.md](https://github.com/slskd/slskd/tree/master#readme) be edited to reflect the need to specify (or benefit of specifying) the user? I used that version and found I was having problems with slskd creating read-only files

PR #1005, opened 2023-12-03
Explain docker --user, add compose examples

andrewjmetzger:
I added some words explaining the `--user`/`user:` specification in Docker, and updated the existing examples to include the relevant options.

I have also added `docker-compose` examples for the sake of completeness.

Closes slskd/slskd#473.


jpdillingham:
Thanks!

PR #1035, opened 2024-01-01
Add configurable limits for number (and size) of queued files, daily and weekly completions

### LINKED ISSUE #388: Add download quotas
#346 introduced slot and bandwidth limiting by group.  Expanding on that idea, file and overall size quotas should be added.

On a per-user basis, track the overall number of files and data downloaded, and prevent users from enqueuing more files if they have exceeded their configured quota for a given time period.  Include settings for daily and weekly limits to start, and consider monthly.

"daily" should be defined as the previous 24 hours, and "weekly" should be defined as the previous 7 days.

This feature depends on #387, as completed transfers need to be persisted so we can count them and aggregate their size.

jpdillingham:
Closes #388


KINGLIFER:
Praying for that lidarr implementation

Issue #1050, opened 2024-01-15
Excessive inotify watch usage/leak


melvyn2:
On both 0.18.2 and 0.19.5, slskd is hitting the inotify user watch limit on my system:
```
INotify Limits:
  max_queued_events    16,384
  max_user_instances   128
  max_user_watches     1,048,576
------------------------------------------------------------------------------
       Pid Uid        App                       Watches  Instances
    541393 992        slskd                    1,046,888          6
...
```
I'm sharing less than 15k files, so even with one watch per file and directory this should not be a problem. This causes problems around the rest of the OS which relies on inotify, like systemctl commands.

melvyn2:
After launch, the watch count sometimes seems to stabilize if it doesn't reach the limit, but still remains very high. Right now I'm looking at 639,179 watches used, which is more than half of the limit.

jpdillingham:
This will be rooted in the dotnet `FileSystemWatcher` that's used under the hood to react to configuration changes; nothing else in the application watches any files.  I can add a setting to disable automatic config reload on change, but with this enabled any config change will require a restart.

Can you share more about how you're running slskd? Every instance has this turned on and if this were a widespread issue I'd expect reports of crashing, so I'm thinking you may have a unique setup.

melvyn2:
I'm running slskd through the nixos module, which results in the following systemd config:
```ini
# /etc/systemd/system/slskd.service
[Unit]
After=network.target
Description=A modern client-server application for the Soulseek file sharing network

[Service]
Environment=<nixos stuff>
EnvironmentFile=/srv/auth/slsk.env
ExecStart=/nix/store/300mj7mnyjk6z3f3h3kxhknp0nqgphxh-slskd-0.19.5/bin/slskd --app-dir /var/lib/slskd --config /nix/store/m8iwprrxqp03m4kw1rzccixszryagaay-slskd.yml
LockPersonality=true
NoNewPrivileges=true
PrivateDevices=true
PrivateMounts=true
PrivateTmp=true
PrivateUsers=true
ProtectClock=true
ProtectControlGroups=true
ProtectHome=true
ProtectHostname=true
ProtectKernelLogs=true
ProtectKernelModules=true
ProtectKernelTunables=true
ProtectProc=invisible
ProtectSystem=strict
ReadOnlyPaths=<path to share 1>
ReadOnlyPaths=<path to share 2>
ReadOnlyPaths=<path to share 3>
ReadOnlyPaths=<path to share 4>
[...]

PR #1062, opened 2024-01-26
Add support for CIDR blacklist files

gsuberland:
Many P2P tools offer support for loading IP range blocklist files, such as those from [i-blocklist](https://www.iblocklist.com/lists). These contain lots of entries, so putting all the CIDRs from one of the lists into the main configuration file quickly becomes unwieldy, and also makes it hard to update the list later.

This PR adds support for loading an external blacklist file, whose entries are merged with any manual entries in the `cidrs` configuration section. Since many blacklist files are quite large (250k+ entries) this PR also introduces a cache for the `IPAddressRange` objects, to prevent having to re-load the file and re-parse every single CIDR string every time a call to `IsBlacklisted` is made. This cache is currently updated at validation time (within the `Validate` function). This works, for now, but a future improvement could be made to detect changes to the blacklist file and invalidate the cache when that occurs, to allow for external updates (e.g. automated blacklist updates) to be reflected at runtime.

I've done my best to follow the style of existing code, but please let me know if there's something that needs tweaking.

### DIFF
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -93,6 +93,7 @@
 #       - <username to blacklist>
 #     cidrs:
 #       - <CIDR to blacklist, e.g. 255.255.255.255/32>
+#     blacklistfile: <path to file containing CIDRs to blacklist>
 #   user_defined:
 #     my_buddies:
 #       upload:
--- a/docs/config.md
+++ b/docs/config.md
@@ -359,7 +359,7 @@ Blacklisted users are prevented from:
 - Retrieving directory contents
 - Enqueueing downloads
 
-Users can be blacklisted by adding their username to the `members` list.  Additionally, users can be blacklisted by IP address, or range of addresses by adding a CIDR entry to the `cidrs` list.
+Users can be blacklisted by adding their username to the `members` list.  Users can also be blacklisted by IP address, or range of addresses by adding a CIDR entry to the `cidrs` list. In addition, an external blacklist file containing CIDRs can be specified using `blacklistfile`, with each CIDR on its own line. Empty lines and lines starting with `#` are ignored.
 
 Users added to the blacklist will be blocked from enqueueing any new files.  Any existing active or queued transfers will need to be cancelled manually.
 
@@ -371,6 +371,7 @@ groups:
       - <username to blacklist>
     cidrs:
       - <CIDR to blacklist, e.g. 255.255.255.255/32>
+    blacklistfile: 'D:\blacklist.txt'
 ```
 
 ## User Defined Groups
--- a/src/slskd/Core/Options.cs
+++ b/src/slskd/Core/Options.cs
@@ -928,9 +928,28 @@ namespace slskd
                 public string[] Members { get; init; } = Array.Empty<string>();
 
                 /// <summary>
+                ///     A pre-parsed cache of IPAddressRange objects describing all blacklisted ranges.
+                /// </summary>
+                private List<IPAddressRange> BlacklistedCIDRs = new List<IPAddressRange>();
+
+                /// <summary>
                 ///     Gets the list of group CIDRs.
                 /// </summary>
                 public string[] Cidrs { get; init; } = Array.Empty<string>();
+
+                /// <summary>
+                ///     Gets the CIDR blacklist file path.
+                /// </summary>
+                public string BlacklistFile { get; init; } = string.Empty;
+
+                /// <summary>
+                ///     Gets address ranges for all blacklisted CIDRs defined by the configuration file and the blacklist file.
+                /// </summary>
+                /// <returns>An IEnumerable of IPAddressRange objects describing the blacklisted CIDRs.</returns>
+                public IEnumerable<IPAddressRange> GetBlacklistedCIDRs()
+                {
+                    return BlacklistedCIDRs;
+                }
 
                 /// <summary>
                 ///     Extended validation.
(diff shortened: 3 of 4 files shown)


jpdillingham:
Thanks so much for the idea and the implementation! I need to investigate a few things but what you've got is a good start.

I think because the `Options` class is used with the underlying "Options pattern" I'm concerned about the framework re-allocating the list when it gets passed around.  For that reason we'll want to find another home for it where it'll be allocated just once.  I'll give it some thought.

In the meantime this exposes a shortcoming of the blacklist functionality; blacklisting a CIDR won't have any effect on search results.  I took a stab at this previously and even left some notes to myself regarding why it wasn't feasible (user IP addresses are not sent along with the search request), but now that I'm thinking about it, It's reasonable to do that check just before sending the results, after the IP address has been resolved.  This occurs in the underlying `Soulseek.NET` library, so I need to do some work over in that repo to enable this.

gsuberland:
Yeah, that definitely sounds like a good thing to move out to its own class if it's being consumed in multiple places.

I realised after sending this PR that it would probably be better to have it support multiple files, rather than just one. The implementation wouldn't really be any more complex, and it would make it far easier to consume and maintain blocklist files that have different purposes and which may be auto-updated from different sources.

It might also be a good idea to store the last modified time of each blocklist file when it is parsed, and re-check that when `IsBlacklisted` is called, to ensure that updates to the blocklist files are reflected at runtime without needing to restart the daemon. The only downside to this approach is that it does force any latency incurred by the re-parse operation to be shunted to the next remote request, effectively blocking it for a short while, but given that this should only affect the first query or connection after the list is updated I suspect it's unlikely to cause issues. The `Clock` class could also be used to check the lists every minute and update when needed, although if we go that route we should be careful to wrap that update operation in a synchronisation primitive to prevent a race condition where a remote connection can come in mid-way through the cache being repopulated, thus bypassing the check.

jpdillingham:
#1064 closes the search response "loophole" that would allow a user that was blacklisted via a CIDR to receive search results.

I agree it would be preferable to support multiple files.  I was thinking the config would be:

```yaml
blacklisted:
  members:
    - <username to blacklist>
  cidrs:
    - <CIDR to blacklist, e.g. 255.255.255.255/32>
  lists:
    - <path to file containing CIDRs to blacklist>
 ```

I also agree about the file refresh.  .NET offers a `FileSystemWatcher` (I believe that's the name anyway) that can watch files for changes and raise events; this is how the config file works under the hood.  I do see multiple events raised on some systems and I have an issue somewhere to try and debounce them, but I'm handcuffed by the internals there.  A 'custom' implementation could definitely be properly synchronized.  A recent issue #1050 where these watchers are causing issues so I'm a little hesitant, maybe the clock would be a safer bet.

Regarding availability during the rebuild, the strategy would be to build a new cache alongside the old and swap them once the rebuild is complete.

I'm going to make a little proof of concept off to the side to see what performance is like and what kind of memory pressure we could expect.  Right now the app "idles" at 12-15mb used and a ~250mb working set.  I'm looking at a ~35mb blocklist and will use that for testing.

I'm also curious what kind of indexing could be used to speed the check up; I'll investigate that more too.

gsuberland:
Running a histogram of CIDR sizes on one of the biggest lists I have (>250k entries) gives me:

- 10.2% are `/32` (one IP)
- 1.8% are `/31` (2 IPs)
- 7.9% are `/30` (4 IPs)
- 40.6% are `/29` (8 IPs)
- 12.7% are `/28` (16 IPs)
- 7.4% are `/27` (32 IPs)

This accounts for over 80% of the IPs in the list.

If we were to expand those subnets out into individual IPs and encode them as `UInt32` values, they could be placed into a `HashSet<UInt32>`, which has extremely fast lookup times.

Ordinarily there would be a memory tradeoff involved with expanding these subnets into individual IPs. However, it isn't quite as simple as that. Each reference type instance on the heap has a type handle pointer, an object header pointer, and an object header which contains a sync block, so on a 64-bit machine that's 24 bytes for the runtime internals. As such, each `IPAddressRange` instance allocates 24 bytes plus the size of the object itself, which has three reference type fields and an `int`, so 28 bytes, making 52 bytes total. Two of those reference fields are typically allocated in the context of slskd's usage, and point to `IPAddress` objects. Each of those contains a `uint`, an `int`, and two reference fields, so that's another 24 bytes (48 bytes including the object header). Based on this, in total, each `IPAddressRange` object theoretically accounts for `52+(2*48)=148` bytes of heap memory consumption. This matches what I see in benchmarks under LINQPad 7 x64.

So, as it turns out, unpacking all CIDRs of size `/32` through `/27` is actually guaranteed to save memory, because a `/32` ends up being 128 bytes. On my large list, these ranges account for 207,836 entries, which turns into 30,759,728 bytes (29.3MiB) once expressed as `IPAddressRange` objects. Expanding these ranges gives 2,083,841 individual IPs, each of which take up 4 bytes, which is 8,335,364 bytes (7.95MiB), a saving of over 70%.

The hashset should also speed things up. I built a little benchmark that works as follows:

1. Load all the CIDRs from my list (>250k) and parse them into `IPAddressRange` objects in a `List<T>`.
2. Randomly select 1000 IPs from within CIDRs in that list.
3. Randomly generate 1000 IPs using random numbers.
4. Join and shuffle both of those into a search candidate array, as `IPAddress` objects.
5. Perform a full GC collection pass and wait for it to complete.
6. Start a stopwatch.
7. Randomly pick an index from 0 to 1999, take the IP from the candidate array at that index, and check for its presence in the CIDR list using `results[i] = cidrs.Any(c => c.Contains(searchArray[idx]));`
8. Repeat step 7 ten thousand times.
9. Stop the stopwatch and print the elapsed time.

I ran the benchmarks five times and got times in the order of 65 to 68 seconds, which means each search through the list is averaging about 6.8ms. I expected it to be slow, but this is _really_ slow!

I then changed the benchmark to unpack all the `IPAddressRange` objects for `/27` CIDRs and smaller into a `HashSet<UInt32>` during the parsing operation, then select all CIDRs larger than `/27` into a `List<T>`. The search candidate array creation was unchanged, to reflect the fact that incoming queries would be in the format of an `IPAddress` object. The check operation was altered to first convert the `IPAddress` object into `uint`, then search the hashset and return true if the item was found, otherwise check the remaining CIDRs in the `List<T>` using the same query logic (`Any` / `Contains`) as the original benchmark.

The hashset approach is _significantly_ faster - reliably around 8-9 seconds. This is a seven-fold performance increase!

jpdillingham:
Thanks for the detailed investigation!  I'll take a crack at it.

I'm also going to take a look at what a solution involving SQLite would look like.  I'm thinking I can add each CIDR to a table with a start and an end (with IPs represented as integers), index both columns, and then use a `BETWEEN` to do the lookup.  It definitely won't be as fast as your approach but it trades memory footprint for disk (+ I/O) which might be a fair trade depending on how much slower it is.

jpdillingham:
@gsuberland I'm finally getting time to work on this a bit, do you mind if i change the target branch and get this merged so I can iterate on it?

gsuberland:
@jpdillingham Sure, no worries :)

jpdillingham:
I explored a few approaches [here](https://github.com/jpdillingham/CIDR-Search) and have a path forward on the approach involving a dictionary keyed on the first octet of each IP covered by a CIDR (since `/7` and lower spans octets) and a list of CIDRs for each.

Detailed benchmarks are in the repo but tl;dr this approach yields an average search time of ~27100ns (0.027ms) while testing 100k random IPs against ~250k CIDRs on a Raspberry Pi 2.  On a ~10 year old PC this improves to ~1600ns (0.0016ms).  In reality IP addresses won't be nearly as random as this test but I'm crossing my fingers the approach holds up.

Unfortunately this will add ~30mb to the memory footprint (assuming the level 1 list from i-blocklist is used) but it's optional so users can decide for themselves.

PR #1065, opened 2024-01-28
Save theme setting to cookie so that it is preserved across page loads

gsuberland:
Saves the light/dark theme state to a "theme" cookie value, which avoids my retinas getting seared like tuna steaks every time I hit F5 😄

### DIFF
--- a/src/web/src/components/App.js
+++ b/src/web/src/components/App.js
@@ -85,6 +85,11 @@ class App extends Component {
 
           await appHub.start();
         }
+
+        const availableThemes = ['light', 'dark'];
+        const preferredThemeRaw = document.cookie.split('; ').map(c => c.split('=')).find(c => c?.at(0) == "theme")?.at(1) ?? availableThemes[0];
+        const preferredTheme = availableThemes.find(theme => theme == preferredThemeRaw) ?? availableThemes[0];
+        this.setState({theme: preferredTheme});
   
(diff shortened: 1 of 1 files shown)


jpdillingham:
Thanks for the PR! I'm a little surprised this isn't persisted, but looking at the existing code it seems to default to the system theme so it's always dark for me and I never really paid attention.

User settings on the UI are something I have yet to think through; the app is built so that someone can swap in their own UI by dropping static files in a folder and changing the `content_path` setting (or just overwriting the contents of `wwwroot`), so whatever the long term solution for UI settings is needs to support this.

That being said, in lieu of an "official" way of persisting UI settings I'd prefer to use `localStorage` since that's what everything else that's currently persisted uses.  I think a good key for this would be `slskd-theme`.

I haven't taken a look at your code in context so apologies if this is accounted for already, but I'd expect the logic to be:

```
if the slskd-theme key exists in localStorage
  apply the stored theme
else
  apply the light or dark theme according to `prefers-color-scheme`
```

This gets a little tricky because the existing logic reacts to the system theme being changed, and I think that'll override the stored setting if/when it happens.  Not a big deal for me, I don't think people change their system theme frequently.  Though maybe according to time of day?  I guess we'll see if anyone complains!

jpdillingham:
Completed by #1201

PR #1080, opened 2024-02-15
[rfc][frontend] Add broader `eslint` config

notjosh:
(PR 1 of 3 related to modernising frontend tooling. See also https://github.com/notjosh/slskd/pull/1, https://github.com/notjosh/slskd/pull/2)

Hiya 👋

This PR introduces [`eslint-config-canonical`](https://github.com/gajus/eslint-config-canonical) as the base codestyle for the project.

⚠️ I haven't tested this branch extensively as it's part of a series to explore the options.

Notes:

- components were renamed from `.js` to `.jsx` so ESLint could apply the correct parser/rulesets, which unfortunately git hasn't kept, which breaks history. We can do the rename as a separate PR to keep continuity.
- some of the linting overrides are pretty arbitrary, but can be re-activated and addressed in followup PRs
- there are a few logic changes to address some specific errors (i.e. extracting `getToken()` out of `session.js` to prevent import loops), but overall shouldn't be too different

Questions:

- are we happy with the code style? (I don't have strong opinions - as long as it's automated + consistent!)
- would one of the existing maintainers prefer to run the `eslint --fix`? It's a minor vanity point, but otherwise my name will appear across the codebase


jpdillingham:
Thanks for putting this up!  I'm scrambling to get a couple of features out on the backend (for the past ~2 weeks 😭) over the next few days and once those are in I'll merge this and the others.

I don't care about git history on the frontend really so no issue with `--fix`.  My only strong opinion about formatting is that I don't like a forced line length period, but if there is one I don't like it to be fewer than 120 characters.  I don't see prettier in here so it's less of a concern.

jpdillingham:
Thanks for your patience! 🚀

PR #1107, opened 2024-03-30
Stop depending on lodash in web app

sethidden:
`lodash` is a CommonJS module, which means that importing just one property from it imports the *whole* library into the bundle that the browser downloads.

Here first I built without lodash (after this PR), and then I built again with current master.

```
> web@0.1.0 build
> craco build

Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  322.28 kB (+24.63 kB)  build/static/js/main.be3e339f.js
  102.95 kB              build/static/css/main.751922b7.css
```

We can see that versus this PR, master is 24 kB heavier, because as mentioned before, it imports all of lodash.

This project uses lodash v4 which is a CommonJS module. However lodash v5 seems to be an ES module, which if importing just one method will import that one method, rather than the whole library.

However since lodash is used just once, I rewrote it to regular JS. If you prefer the ease of just importing orderBy it's ok to update to lodash v5 instead, just let me know.

First introduced https://github.com/slskd/slskd/commit/c9faf80c4988ef52fb2047763f455e7c30f4bbaf


jpdillingham (approved):

jpdillingham:
@sethidden LGTM! I'll merge it once CI is done.

Issue #807, opened 2022-12-29
How to download all folders in a folder?


verbes4:
When browsing a users files, I can't seem to download a folder which contains exclusively more folders, with files inside them. 
For example:
![image](https://user-images.githubusercontent.com/24574201/210006774-0ac41567-ba20-4aeb-a23f-4fc2884eca51.png)
^ I want to download all folders in the "Kevin MacLeod" folder I've selected, but there is no download button. 

![image](https://user-images.githubusercontent.com/24574201/210006923-cec46630-a538-4d4f-a92e-a8b2f34c0b05.png)
^ Only when I click into the folder within and select all the files does the download button appear. I understand I can do this for each folder I want to download, but if I could select them all by just choosing the top level folder this would be preferred.

jpdillingham:
Unfortunately this isn't supported at the moment, but I tweaked this issue to capture the feature.

I don't think it would take a lot of work to implement as the data is all there in the browser and there's already logic to batch single folders together.

jpdillingham:
Adding `help wanted` and `good first issue` labels; this would be a UI-only change if anyone wants to take a shot at it.

At a high level, this requires:

* Adding a button/icon to each folder in the result window
* Adding a modal that would open and recursively list the contents of the selected folder and prompt the user to confirm/cancel
* Make an API call to enqueue the downloads, passing the contents of the modal as an array

Bladdermate:
I'd really appreciate this too! Thanks for the great job with this app!

KyokoMiki:
> Adding `help wanted` and `good first issue` labels; this would be a UI-only change if anyone wants to take a shot at it.
> 
> At a high level, this requires:
> 
> * Adding a button/icon to each folder in the result window
> * Adding a modal that would open and recursively list the contents of the selected folder and prompt the user to confirm/cancel
> * Make an API call to enqueue the downloads, passing the contents of the modal as an array

I've been troubled by the issue that slskd does not support recursively downloading folders from the browse interface, but my C# programming skills are virtually non-existent. Therefore, I'm wondering if I could make modifications through ChatGPT to ultimately add this feature to slskd.

I hope to find out where the code for 'Adding a button/icon to each folder in the result window' and 'Make an API call to enqueue the downloads, passing the contents of the modal as an array' is located (which file and where exactly...?). I'll take a look with ChatGPT and try to add this feature. I would be very grateful if this is possible.

Dolfost:
[...]

PR #1113, opened 2024-04-11
Add the ability to recursively download folders from the browse interface

KyokoMiki:
#807 

a6d57e3914d425108532f63769c02f3e

45cdf41cb4ab9d779c61f418ab2c21bb

278c6defcd355986930005338be85c9b

Previously, the entire backend download directory was hard-encoded and could not be manually specified. Therefore, I modified api endpoint:

```
namespace slskd.Transfers.API
{
    public class QueueDownloadRequest
    {
        public string Filename { get; set; }
        public long Size { get; set; }

        public string Path { get; set; }
    }
}
```

I added the ability to specify the download directory (relative path). However, because I modified slskd.Transfers.Transfer, it **caused the transfers.db database to no longer be compatible with the previous version**. The "Path" column needs to be manually added to the transfers.db database.

It should also be possible to use a method similar to subnet masks, adding a small segment after the filename to indicate where the relative path starts. However, I believe manually specifying the download file name is still quite important. In the future, there may be other issues that require this functionality. Therefore, I wanted to submit this first and ask for the developer's opinion.

### DIFF
--- a/src/slskd/Common/CommonExtensions.cs
+++ b/src/slskd/Common/CommonExtensions.cs
@@ -352,10 +352,42 @@ namespace slskd
         /// </summary>
         /// <param name="remoteFilename">The fully qualified remote filename to convert.</param>
         /// <param name="baseDirectory">The base directory for the local filename.</param>
+        /// <param name="relativePath">Optional relative path to override the default path structure.</param>
         /// <returns>The converted filename.</returns>
-        public static string ToLocalFilename(this string remoteFilename, string baseDirectory)
+        public static string ToLocalFilename(this string remoteFilename, string baseDirectory, string relativePath)
         {
-            return Path.Combine(baseDirectory, remoteFilename.ToLocalRelativeFilename());
+            // Check if filename contains custom path information (format: "originalFilename|customPath")
+            string actualRemoteFilename = remoteFilename;
+            string customPath = relativePath;
+
+            if (remoteFilename.Contains("|"))
+            {
+                var parts = remoteFilename.Split('|', 2);
+                if (parts.Length == 2)
+                {
+                    actualRemoteFilename = parts[0];
+                    customPath = parts[1];
+                }
+            }
+
+            if (string.IsNullOrEmpty(customPath))
+            {
+                return Path.Combine(baseDirectory, actualRemoteFilename.ToLocalRelativeFilename());
+            }
+            else
+            {
+                // For custom paths, we need to sanitize the path and ensure proper directory separators
+                var sanitizedCustomPath = customPath.LocalizePath();
+                var parts = sanitizedCustomPath.Split(Path.DirectorySeparatorChar);
+
+                for (int i = 0; i < parts.Length; i++)
+                {
+                    parts[i] = parts[i].ReplaceInvalidFileNameCharacters();
+                }
+
+                var finalPath = string.Join(Path.DirectorySeparatorChar, parts);
+                return Path.Combine(baseDirectory, finalPath);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


KyokoMiki:
> Thanks for the contribution! I took a look at the changes in the UI and I think they look great!
> 
> I glanced at the backend changes and I'll need to do a more thorough review; path normalization has been contentious at several different points during development and I want to make sure this isn't going to introduce any regressions. I didn't expect that there would need to be a backend change for the application to support this.

Thank you very much!

This issue is because the backend directly specifies the last two parts of the remote directory as the relative path when handling downloads (a/b/c/x.flac -> c/x.flac). Therefore, apart from modifying the API to specify downloading to b/c/x.flac and adding the ability to override paths in the backend, I couldn't come up with a better solution. My personal ability is limited, so there might be better approaches.

lazee486:
just discovered soulseek, and the webui is awesome, hope this feature comes soon :) serious thanks!

ricardopeters-dev:
Hi! Thanks for the work on this, just wanted to explain an issue I’m running into with multi-disc albums.

Since it's not yet possible to download folders recursively, I have to manually download each subfolder (like CD1, CD2). The problem is they all get saved into the same top-level folder - for example, /downloads/CD1 - so when I grab another album with the same disc structure, the files end up mixed together or overwritten.

For example, downloading The Beatles – White Album and Smashing Pumpkins – Mellon Collie and the Infinite Sadness results in both albums' CD1 and CD2 tracks ending up in the same place.

It’d be great if full folder structures could be preserved, or if recursive downloading was supported to avoid this kind of conflict.

Appreciate all the progress so far!

KyokoMiki:
@jpdillingham I have refactored the recursive download functionality. The implementation in this version is much simpler than before and will no longer produce breaking changes like it used to.

jpdillingham:
The backend changes aren't necessary, this can make a single request and include all of the files with their full path, the same as what's done when selecting files and clicking the download button currently.

Where the files go on disk is a separate issue that I'm still thinking about a solution for.

zoeleu:
any reason why this hasn't been merged yet?

Tak-MK:
mini bump here (?) 😄

KyokoMiki:
@jpdillingham I refactored this feature based on the latest backend. I've tested it and verified that it works properly. I noticed that there is no longer a need to modify the backend, and the recursive download functionality can be implemented simply by modifying the frontend. Additionally, I have also added a recursive download button to the search interface.

Jo3l:
We are waiting for this feature!

jpdillingham:
FYI, when reviewing the latest changes I started to get a bit uncomfortable with the idea of users blindly enqueuing top-level directories without having a sense of what is in them (unless they looked, in which case they would not need this feature).

I'm iterating on a design that will display expandable directories in the file list, and that will give users an icon they can click to navigate to that directory in the tree to further inspect the contents.  Totals for files and directories are shown, and if a user selects directories the download button will display the total file count and size so there should be no surprises.

Here's a sample:

image

In the future I might expand this view to allow for recursive drilling into directories, but for right now I want to limit the amount of effort that goes in because the entire approach to browsing needs to be reworked as documented in #317

KyokoMiki:
It looks like we'll need another two years before we can use the recursive download feature you envisioned for perfectionists. Since you can't see the demand for this feature that exists around the world, and you're unlikely to experience the frustration I feel when using your downloader—when two subfolders both named 'Scans' get downloaded together and become impossible to separate—then you can keep playing with it yourself. I'm going to go contribute to your project's competitors instead.

jpdillingham:
I'm not competing with anyone, I'm not going to accept changes that I'm not entirely happy with, and I certainly don't owe you or anyone else anything.

Also, #1746 adds the ability to control download location and it will be included with the next release.  Along with the change I've shown above.

zoeleu:
@jpdillingham Does #1746 add batch downloading?

jpdillingham:
@zoeleu  The concept of a batch and some related features have been added in #1664 and #1720,  More functionality will be added later.

PR #1137, opened 2024-05-31
Add option to disable console log colours

zliebersbach:
I require this option due to my container manager not parsing console colour codes correctly, and giving me a bunch of trash:

Screenshot 2024-05-31 at 15 31 18

--------

**Note**

I have implemented this boolean config option the same way as the disk option, except I noticed that the environment variable and argument options do not properly 'invert' the value as it should. i.e. `--no-disk-logger` should set `LoggerOptions.Disk = false`, but it doesn't actually do anything.

I've simply left it as is, since it works as expected through the YAML config file.


jpdillingham:
I fixed the disk logging options in #1139 and in doing so remembered that any boolean options have to be expressed as a negative because the command line implementation can't negate booleans; if the flag is present it's `true`, not present its `false` (this is a shortcoming of the implementation)

I think if we were to move forward with this option the backing property would need to be defaulted to `false` and the name would need to be something like `NoColors` or `Colorless` so the semantics would be correct.

All that being said, I'm pretty sure we can find a way to disable the colors automatically; they're only useful when a human (probably a dev) is looking at the console and there's probably something under the hood we can look at to tell.

jpdillingham:
Doing some digging and I see that Serilog is _supposed_ to be able to identify these situations and suppress the theme: https://github.com/serilog/serilog-sinks-console/issues/50

The `applyThemeToRedirectedOutput` option is `false` by default, which leads me to believe that something about your hosting environment is keeping the underlying `IsOutputRedirected` flag set.

Reading through the linked issue I see that people do want the colors in these scenarios, so I think we're back to letting the user decide.

zliebersbach:
Great @jpdillingham,

I have honestly not had good experience with Serilog correctly detecting whether or not to apply themes to the output. I think the right choice here is to have a user option.

Let's rename the property to `NoColor`, which seems to be _some_ sort of accepted standard: https://no-color.org

Perhaps it's not worth having it in the config, and simply reading the `NO_COLOR` env variable directly?

jpdillingham:
> Perhaps it's not worth having it in the config, and simply reading the NO_COLOR env variable directly?

Certainly would have been a shorter path!  I ended up adding an explicit check for it so that it'll work automatically on systems where this is set.

All of the `SLSKD_` environment variables are populated using Microsoft's `IConfiguration` stuff, which in retrospect is a bit more complicated than what was probably needed.

PR #1157, opened 2024-06-27
Improve reverse proxy setup for Nginx

milleniumbug:
Fixes the problem I encountered with the reverse proxy setup when I tried out downloading a 5GB file from my slskd setup.

By default Nginx has its own limit of body size of 1MB, and also its own proxy buffering. These two settings disable them.


jpdillingham (approved):
thanks!

Issue #320, opened 2022-01-03
Investigate issue with shares configured via symlink


jpdillingham:
Reported in #155 

It seems symbolic links with .NET Core might be a bit of a contentious issue, and probably doesn't work as transparently as I would have assumed.  I'll need to try configuring a share and see what happens.  There may be a way to differentiate directories from symlinks and resolve them differently.

ppom0:
I think I have this issue. It seems like it shares the file, but only the first bytes, same length of the symlink itself. Not sure though, it's hard to debug.

jpdillingham:
@ppom0 FYI this is fixed in #1165

PR #1158, opened 2024-06-28
Handle file symlinks in shares correctly

milleniumbug:
#320 

`new FileInfo(fileName).Length` reports the filesystem length of a symlink itself, which is a nonsensical value.

If a file happens to be shorter than that value, the download never finishes and client disconnects through a timeout.

If a file happens to be longer than that value, the downloaded file is truncated to the length equal to that value.

This pull requests adds code that resolves the symlink first (if it's a symlink).

I have tested the code on my Linux machine.

### DIFF
--- a/src/slskd/Files/FileExtensions.cs
+++ b/src/slskd/Files/FileExtensions.cs
@@ -59,4 +59,25 @@ public static class FileExtensions
 
         return (UnixFileMode)mode;
     }
+
+#nullable enable
+    public static FileInfo? TryFollowSymlink(this FileInfo fileInfo)
+    {
+        try
+        {
+            return fileInfo.FollowSymlink();
+        }
+        catch (IOException)
+        {
+            return null;
+        }
+    }
+
+    public static FileInfo FollowSymlink(this FileInfo fileInfo)
+    {
+        FileSystemInfo fileSystemInfo = fileInfo;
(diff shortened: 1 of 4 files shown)


jpdillingham:
thanks for putting this up! i have some questions about how this works for shares as well, so i'll need some time to pull it down and test it out.

milleniumbug:
Sure, if you have a specific case in mind I can try making sure it works. This PR addresses the case where a shared directory contains a symlink to a file. The case where a shared directory contains a symlink to a directory seems to work already. I didn't however test the case where the entry in the list of shares configuration refers to a symlink itself. I also didn't test the Windows symlink capability (Windows is different in that it has "file symlinks" and "directory symlinks", as in the type of the thing it's referring to is encoded in the symlink itself).

jpdillingham:
I'm going to merge this into a different branch so I can mess around with it.  I want to be cautious because cross-platform file I/O has been pretty contentious in the past and I want to make sure the solution works transparently across all scenarios (you brought up a great one regarding the newer Windows symlinks)

I'm mainly thinking about how metadata is derived during the share scan, which requires that file contents are read.  I've been somewhat surprised that it doesn't crash; I have no idea what info is returned.  It's possible that the library I'm using that follows symlinks internally.

Issue #1126, opened 2024-05-11
Non-media files display "Invalid Date'


jpdillingham:
Not sure when this started but I just noticed today.  Maybe a dependency update did it?

Either way, should be an easy fix.

PR #1159, opened 2024-06-30
Fix Non-media files display "Invalid Date'

milleniumbug:
#1126


jpdillingham:
thanks again!

closes #1126

PR #1165, opened 2024-07-08
Add support for file symlinks

### LINKED ISSUE #320: Investigate issue with shares configured via symlink

jpdillingham:
Builds on #1158 by @milleniumbug.

Closes #320 

The logic is encapsulated in the `FileService` class along with the other file I/O, as cross-platform support has been shaky in the past and I don't trust it fully.  Symlink support in .NET is fairly new and I was able to find at least 3 open issues in the dotnet runtime repository related to erroneous results, and I suspect issues will be encountered in the future.

I tested this with both directory (which were working already) and file symlinks, confirming that both browse and search responses return the correct file size, and that downloads of each complete as expected.  I haven't tested with the relay, but there's no reason it should be any different than regular uploads.

Moving forward (and this is more of a note to myself I guess), no application logic should attempt to resolve `FileInfo` the conventional way; `new FileInfo(<path>)`.  Instead it should use the `ResolveFileInfo()` method from the `FileService` class, which navigates symlinks defensively.

### DIFF
--- a/src/slskd/Application.cs
+++ b/src/slskd/Application.cs
@@ -40,6 +40,7 @@ namespace slskd
     using Serilog.Events;
     using slskd.Configuration;
     using slskd.Core.API;
+    using slskd.Files;
     using slskd.Integrations.Pushbullet;
     using slskd.Messaging;
     using slskd.Relay;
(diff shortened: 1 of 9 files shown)


jpdillingham:
@milleniumbug FYI; you had the right solution in the relevant places, I just went through and made it bulletproof (🤞) and applied the change everywhere.

PR #1168, opened 2024-07-14
fix(web:system:shares): add `theme` to function args

### LINKED ISSUE #1167: Theme not applied to the Shares/files modal.
image

this happens because we arent passing theme to  `ContentsModal`

https://github.com/slskd/slskd/blob/26ecbb6955d9281bf94af4df4d55788cd05a0211/src/web/src/components/System/Shares/index.jsx#L155-L157

I have a fix PR ready.

MagneticNeedle:
fixes #1167 , 

also can we pls work a bit on dev docs :face_holding_back_tears:

### DIFF
--- a/src/web/src/components/System/Shares/index.jsx
+++ b/src/web/src/components/System/Shares/index.jsx
@@ -33,7 +33,7 @@ const CancelButton = ({ cancel, working }) => (
   </ShrinkableButton>
 );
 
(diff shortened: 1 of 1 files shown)


jpdillingham:
Thanks for this, I noticed the other day but forgot to create an issue.

Can you elaborate on the dev docs? What do you feel is lacking?

PR #1196, opened 2024-09-02
fix(docker): Allow container to be ran as read-only filesystem

onedr0p:
This is env var is required for those that have a strict security policy on running containers. I don't see a downside to enabling this otherwise and it seems like a common issue with dotnet based apps running in containers.

More info here

https://github.com/Radarr/Radarr/issues/7030#issuecomment-1039689518
https://github.com/dotnet/runtime/issues/9336
https://github.com/recyclarr/recyclarr/pull/231

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,8 @@ VOLUME /app
 
 HEALTHCHECK --interval=60s --timeout=3s --start-period=5s --retries=3 CMD wget -q -O - http://localhost:${SLSKD_HTTP_PORT}/health
 
-ENV DOTNET_BUNDLE_EXTRACT_BASE_DIR=/.net \
+ENV COMPlus_EnableDiagnostics="0" \
+  DOTNET_BUNDLE_EXTRACT_BASE_DIR=/.net \
   DOTNET_gcServer=0 \
   DOTNET_gcConcurrent=1 \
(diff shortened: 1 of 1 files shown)


jpdillingham:
Thanks!

My only reservation is that it appears disabling this breaks profiling tools (https://github.com/dotnet/runtime/issues/96227), and I'm not sure to what extent the embedded Prometheus metrics rely on profiling to work.

Thumbing through the [documentation](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#eventpipe-variables), my best guess is that the functionality disabled by this envar that's conflicting with the read only filesystem would be the EventPipe.  If this setting breaks Prometheus I can try fiddling with the more granular settings.

In case that happens, can you share a reproduction of the issue with the current container?  Or is it as simple as passing `--read-only` to `docker run`?

onedr0p:
I have this env set on my deployment of slskd and the prometheus endpoints still works and has data, wether or not it is has all the data is another question 😄 

I think passing read-only to docker would reproduce this issue, I have this deployed in k8s however with the following security context

```yaml
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL
          readOnlyRootFilesystem: true
```

Issue #1198, opened 2024-09-04
[Feature Request] Add Configuration for default dark/light mode in web UI


nhawke:
Right now, slskd's web interface always starts in the device theme (Dark theme on my machine). I prefer light theme in slskd, but because of my device default, whenever i log in again or reload the page, it's back to dark. There should be an option to configure the default theme. Possibly in the config yaml as:

```yaml
web:
   theme: 'dark'
```

or as a boolean

```yaml
web:
  darkTheme: false
```

jpdillingham:
Good idea, thanks!

nhawke:
Since opening this issue, I created PR #1201, which addresses my issue with the theme not sticking to what I'd prefer. This might now be somewhat redundant / not worth the hassle.

mathiascode:
Closing due to comment above.

PR #1201, opened 2024-09-06
Remember theme setting between page loads

nhawke:
Previously, the theme setting would default to the device preference on each load. So, if a user preferred light theme for slskd despite having a dark device theme, they would need to toggle the theme on each page load. By saving the theme preference in localStorage, we can override the device default with the user's manual preference.

Addresses my main concern in issue #1198.

This is a re-implementaiton of the (abandoned?) PR #1065 but implements the feature using `localStorage` per the suggestion in [the comment on that PR](https://github.com/slskd/slskd/pull/1065#issuecomment-1913637361).

### DIFF
--- a/src/web/src/components/App.jsx
+++ b/src/web/src/components/App.jsx
@@ -149,8 +149,8 @@ class App extends Component {
           'change',
           (event) => event.matches && this.setState({ theme: 'light' }),
         );
-
     }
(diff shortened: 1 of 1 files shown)


jpdillingham:
Looks (and works) great, thanks!

Issue #413, opened 2022-04-09
Roadmap


jpdillingham:
# [0.16: Improved Search Experience](https://github.com/slskd/slskd/milestone/2)

The current search experience restricts users to one search at a time, doesn't display results until the search is complete, and doesn't give the user any control over when a search is determined to have been completed.

This milestone is dedicated to a ground-up refactor of the search functionality to resolve these shortcomings, both to improve overall user experience and to enable search automation, for which concurrent searches are critical.

# [0.17: Guided Search](https://github.com/slskd/slskd/milestone/3)

Presently, users enter a simple search phrase and must review results to determine whether they are suitable. For example, when searching for a full album, the user must know the track listing ahead of time, and must compare results against the listing while looking for missing files, etc.

The Guided Search feature will allow users to select an artist and an album to search for, and will use data from MusicBrainz (and potentially Discogs) to rank and automatically select the best results as determined by user-defined criteria.

# [0.18: Post-Download Automation](https://github.com/slskd/slskd/milestone/4)

When downloads finish, users may want the option to perform various tasks; moving files between network shares, setting file permissions, post-processing metadata with a program like [beets](https://beets.io/), or notifying external applications.

The Post-Download Automation milestone will introduce a suite of tools and options to give users the ability to accomplish these tasks, and more.

jpdillingham:
I've removed the min-milestone 0.16.x (misc cleanup and quality of life improvements) as the biggest chunks of that are now complete.

Based on feedback I've gotten from the community, I've moved the Guided Search milestone from 0.17 to 0.19 and introduced the new 0.17 milestone Share Scanning at Scale

Effectively this pushes Lidarr integration out again, but in the spirit of Agile software development I need to work on the most important things first.

jpdillingham:
I've just updated the roadmap again.  Changes:

* Added the 1.00 milestone
* Removed numbering from Post-Download Automation and Guided Search milestones

Currently the versioning scheme (aligned with numeric milestones) only makes sense to me, and I don't even get much benefit out of it, so I want to move to semver.  I had always planned to start using semver after I had gotten to 1.00, and I believe the app is near or perhaps past 1.00 at this point.

This pushes download automation out again, and again it is because people are asking for things other than download automation.  This does remain a personal goal, so it will be done eventually even if it is out-of-band.

aopom:
Many thanks for this transparency, it's super cool to know where the project is going! Post-Download Automation will be a great feature to have. Exciting!

[...]

Issue #348, opened 2022-02-06
Ability to run a script on download completion


tandy-1000:
I currently use [betanin](https://github.com/sentriz/betanin) with qbittorrent to automatically send downloads to beets, streamlining the import into my library, it would be nice to have this functionality.

jpdillingham:
Thanks for the suggestion! I think this is fairly easily doable, I just need to think of the interface.

Taking a look at betanin specifically, it looks like it wants to process an entire directory at the same time, which would be difficult to pull off, because Soulseek doesn't have a concept of albums or directories, just single files.  This is an issue that is holding back Lidarr integration as well.

I'll have to think of the best way to introduce functionality that groups files as an "album", perhaps a new button on the UI like "Download as Album" or something.

jpdillingham:
To elaborate on the sticking point, acting on the completion of a single download is very straightforward; that's how the FTP integration works today.  Acting on the completion of a directory/album is more difficult because of a loose association between individual file downloads.

I think I probably need to introduce a way to download an entire directory; this would give me the ability to act when all of the files within it complete.

Unfortunately a feature like this will have a large degree of ambiguity due to the fact that some users don't arrange their shared directories as full albums.  Many do, but the outliers will cause this functionality to break.

I think over time this can be improved, especially when I (finally) implement the feature(s) that allow searching by MusicBrainz/Discogs IDs, and can therefore ensure that directories contain a full album before enqueueing them.

tandy-1000:
That makes sense, I think "Download directory" probably makes more sense for most use cases until we have the musicbrainz integration, but yeah that makes sense about some directories being a mess..

boyska:
Isn't it enough to have this logic:

 - WHEN a file has been downloaded
 - IF there is no other file being downloaded in the same directory
 - THEN run the script

?

While I agree that different users use directories differently, I see no problem in allowing a script being invoked when a directory has been completed. Users who prefer to download everything in the same directory will probably not find this script very useful, but that's quite obvious. The only thing that matters is set a clear expectation of the behavior, and clarify that this has no notion of _Album_, only _directory_.

jm2c

jpdillingham:
Released in [0.22.0](https://github.com/slskd/slskd/releases/tag/0.22.0)

@boyska I forgot to thank you for the suggestion; the implementation is just as you described.

boyska:
> Released in [0.22.0](https://github.com/slskd/slskd/releases/tag/0.22.0)
> 
[...]

Issue #94, opened 2021-06-09
Add webhook support


jpdillingham:
To enable interop with other applications, design and implement a system for sending event-driven data out of slskd via webhooks.

This will involve the creation of an in-memory event bus and the dispatch of events from both the Soulseek client and the slskd application.  Subscribed to the event bus will be a handler that examines events and calls the appropriate webhook along with the event payload.

Webhook configuration should allow users to specify a URL for the webhook and a list of events they want to subscribe to.

Some example use cases:

* Completed transfers (for file handling automation)
* Received private messages
* Server disconnect/reconnect
* Diagnostics (performance metrics, error logs, etc)

jpdillingham:
Released in [0.22.0](https://github.com/slskd/slskd/releases/tag/0.22.0)

PR #1215, opened 2024-09-15
Add 'Events'

jpdillingham:
A long-standing [goal](https://github.com/slskd/slskd/issues/413) of this project is to support various automation features, including (but not limited to), yes, Lidarr integration.  A necessary stepping stone for that support is to establish a mechanism for signaling when various events occur so that external applications can be notified.

This PR establishes an internal 'event bus' to provide a foundation for that signaling, and implements the first two events:

* `DownloadFileComplete`, raised when an individual file download finishes
* `DownloadDirectoryComplete`, raised when the download of the final file in a directory finishes

Events are persisted to a new `events` database and can be retrieved via the UI and API via the `/events` route.

Up next will be implementation of webhook support (#94), and the ability to run a shell script (#348) when events are raised.

## Testing

* [x] An event record is created for each and every file downloaded
* [x] An event record is created for each directory downloaded, upon completion of the final file in the directory

## Out of scope

To be included shortly:

* Integrations that depend on events (e.g. shell scripts, webhooks)


config/slskd.example.yml
```diff
@@ -168,7 +168,8 @@
 #   files:
 #     complete: 20160 # 2 weeks
 #     incomplete: 43200 # 30 days
-#   logs: 259200 # 180 days
+#   events: 30 # days
+#   logs: 180 # days
```
jpdillingham, line 172: the `logs` option was incorrectly specified in something other than days

PR #1219, opened 2024-09-22
use DOTNET_ var in docker

kasperk81:
.net 6 onwards, DOTNET_ is preferred over COMPlus_ https://github.com/dotnet/runtime/commit/fbdcec9060f46353761ac8ee14d11d6206049395 it first uses DOTNET_, and then COMPlus_ as a legacy fallback.


jpdillingham:
Thanks for this, I never would have known!

This setting was introduced in #1196 to enable support for running the container on a read only filesystem; I'll take some time this weekend to test and make sure this doesn't introduce a regression.  Let me know if you've done so already.

jpdillingham:
Thanks for linking that, sheds more (easily digestible) light on why this conflicts with read only configurations.

Thanks again for the contribution!

PR #1242, opened 2024-11-16
docs: fix YAML indentation

ppom0:
Hi, I think there is a typo in the indentation of this YAML block. Do you agree?


jpdillingham:
Good catch, thanks!

PR #1249, opened 2024-11-19
Update reverse_proxy.md

lemonade-soda:
Added instructions for reverse proxy via SWAG with subdomain, as well as a sample config file.


jpdillingham:
Thanks!

PR #1251, opened 2024-12-06
refactor: Update Dockerfile and Entrypoint script

vitoyucepi:
## Summary

This pull request updates the infrastructure used for building container images.
1. The Dockerfile has been moved to the containers directory, and the CI workflow file has been updated to reflect this new location.
1. A new entrypoint script has been created to handle the container's startup process.
1. The entrypoint script now changes the user for the main application.

## Details

The container image build infrastructure should be in the `containers` directory. All files needed exclusively for building images should be placed within this directory. The contents of the `containers/rootfs` directory will be copied to the root directory (`/`) of the container, allowing any files to be delivered to their exact locations inside the container.

The previous entrypoint script chains `tini` -> `sh` -> `slskd`. The updated entry point script changes the user before starting `slskd` and directly transfers control to the `slskd` binary. The `app` user is provided by the `dotnet` image itself.
This change **may break** startup process for the users because shared files and directories were owned by root. I can add `chown` calls inside the entrypoint.

## Additional

- The `HEALTHCHECK` now uses `curl` instead of `wget` for improved error messages. 
- Reformatted the GitHub workflow script using Prettier for consistency.  
- I adopted the `RUN` command chaining style, similar to official Docker images.  
- Created a symlink from `/slskd/slskd` to `/usr/bin/slskd`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,7 @@ jobs:
             ghcr.io/slskd/slskd:canary
           cache-from: type=gha
           cache-to: type=gha,mode=max
+          file: containers/Dockerfile
 
       - name: Build and push Release
         if: startsWith(github.ref, 'refs/tags/')
@@ -170,6 +171,7 @@ jobs:
             ghcr.io/slskd/slskd:latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
+          file: containers/Dockerfile
 
   release:
     name: Create Release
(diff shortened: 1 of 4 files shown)


jpdillingham:
Thanks for the PR but I'm not a fan of this file placement

vitoyucepi:
Hi @jpdillingham,
Will it make sense to change the PR, or is it just no?

PR #1256, opened 2024-12-16
Added UploadFileCompleteEvent

samhatter:
## Summary
This PR adds and implements an event, UploadFileCompleteEvent, similar to DownloadFileCompleteEvent.

## Details

- Adds EventBus to UploadService
- Creates UploadFileCompleteEvent type
- Raises a UploadFileCompleteEvent on upload completion

### DIFF
--- a/src/slskd/Events/Types/Events.cs
+++ b/src/slskd/Events/Types/Events.cs
@@ -26,6 +26,7 @@ public enum EventType
     Any = 1,
     DownloadFileComplete = 2,
     DownloadDirectoryComplete = 3,
+    UploadFileComplete = 4,
     Noop = int.MaxValue,
 }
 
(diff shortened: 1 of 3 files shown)


src/slskd/Events/Types/Events.cs
```diff
@@ -50,11 +51,20 @@ public sealed record DownloadDirectoryCompleteEvent : Event
...
     public string Username { get; init; }
 }
 
+public sealed record UploadFileCompleteEvent : Event
+{
+    public override EventType Type { get; } = EventType.UploadFileComplete;
+    public override int Version { get; } = 0;
+    public required string LocalFilename { get; init; }
+    public required string RemoteFilename { get; init; }
+    public required Transfer Transfer { get; init; }
```
jpdillingham, line 65: thanks for adding `required`, i keep forgetting about it

src/slskd/Transfers/Uploads/UploadService.cs
```diff
@@ -24,13 +24,15 @@ namespace slskd.Transfers.Uploads
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Diagnostics.Eventing.Reader;
```
jpdillingham, line 27:
```suggestion
    using System.Diagnostics.Eventing.Reader;
```

this shouldn't be necessary.  if you're using VS Code it may have added it automatically

jpdillingham:
Thanks for this! I just had a comment about (what I hope is) an unused `using`, but other than that it looks good.

samhatter:
Thank you,

I think that might have been erroneously added. I will make the suggested
changes. I am fixing my repository at the moment so give me a second.

Sammy

src/slskd/Transfers/Uploads/UploadService.cs
samhatter: Fixed. Thank you.

jpdillingham:
Thanks again!

I'm planning to do a release tonight or tomorrow, so this should be out shortly.

samhatter:
Amazing!

Thank you,
Sammy

PR #1277, opened 2025-01-15
feat(container): Add `jq`

onedr0p:
Hi 👋🏼 

For the script integrations since the $DATA is in JSON it would be nice to write our custom scripts a bit cleaner and using `jq` would allow for that.

So instead of:

```sh
name=$(echo "$1" | awk -F'"localDirectoryName": "' '{print $2}' | awk -F'",' '{print $1}')
```

we can use:

```sh
name=$(jq -r '.localDirectoryName' <<< "$1")
```

Besides adding a few kilobytes to the image I don't see any harm with including this tool by default?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,7 @@ ARG REVISION=0
 ARG BUILD_DATE
 
 RUN apt-get update && apt-get install --no-install-recommends -y \
+  jq \
   wget \
   tini \
   && \


onedr0p:
This isn't every tool, it's just jq and it would help with wiring up automation for the scripts integration.

davidarkemp:
I appreciate that, but if you start adding "just one more tool", where do you stop? See the discussion about VPNs for example.
The point about OCIs is that they're layered, so customising a base image to your liking is pretty trivial.
If you're using docker compose, you can update your yaml to point the Dockerfile instead of the image.
If you using the CLI directly, then something like `docker run $(docker build -q .)` would do it.

onedr0p:
> I appreciate that, but if you start adding "just one more tool", where do you stop?

At jq? I dunno what to say except the script integration sends json to the custom script you specify so having a tool to easily parse the output makes sense.

> The point about OCIs is that they're layered, so customising a base image to your liking is pretty trivial. If you're using docker compose, you can update your yaml to point the Dockerfile instead of the image.

I don't need the lecture on building custom images, I have been [building containers for many projects](https://github.com/onedr0p/containers) for awhile, but these are mainly for apps that do not have an upstream container present.

> See the discussion about VPNs for example.

I don't even need to look that up to understand that the image should never have that kind of tooling packaged.

All I am saying it sometimes it's nice UX to include some basic CLI tooling, for example [here](https://github.com/actions/runner/pull/3056).

ishioni:
I think jq would be useful since the app already outputs json. Parsing json with awk is suboptimal, and jq would add negligible size to the image.

jpdillingham:
I agree jq would be useful, I'm waiting to see how the script functionality plays out and want to see the resulting image size before merging this.  I don't think it has any dependencies so I'm guessing it's negligible.

I also agree about this being a slippery slope of sorts; I wouldn't want a bunch of tools in here solely to support scripting.  It makes sense to include this since the app only passes json and therefore every scripting use case would require jq (or fiddling with awk)

jpdillingham:
Current size: 
image

jpdillingham:
image

Added less than 500kb

Thanks @onedr0p!

PR #1290, opened 2025-02-13
Fix log retention to be days in docs

SpyrosRoum:
Assuming the example config and the note a few lines above this change are correct, this should be updated


jpdillingham:
You're right, it's in days.  Thanks!

PR #1292, opened 2025-02-16
Refactor user-defined scripts

### LINKED ISSUE #1278: Script integration not working
Hi 👋🏼 

I am having trouble getting the scripts integration to work, I have tried the below and also tried with a custom script but I cannot get it to run.

## config

```yaml
integration:
  scripts:
    debug:
      on:
        - DownloadFileComplete
        - DownloadDirectoryComplete
      run: echo "$DATA" >> /config/debug.txt
```

## logs

```
[17:02:33 WRN] Failed to run script 'debug' for event type DownloadFileComplete: Broken pipe
System.IO.IOException: Broken pipe
 ---> System.Net.Sockets.SocketException (32): Broken pipe
   at System.Net.Sockets.Socket.Send(ReadOnlySpan`1 buffer, SocketFlags socketFlags)
   at System.IO.Pipes.PipeStream.WriteCore(ReadOnlySpan`1 buffer)
   --- End of inner exception stack trace ---
   at System.IO.Pipes.PipeStream.WriteCore(ReadOnlySpan`1 buffer)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.IO.StreamWriter.WriteLine(String value)
   at slskd.Integrations.Scripts.ScriptService.<>c__DisplayClass10_1.<HandleEvent>b__3()
[17:02:33 WRN] Failed to run script 'debug' for event type DownloadDirectoryComplete: Broken pipe
System.IO.IOException: Broken pipe
 ---> System.Net.Sockets.SocketException (32): Broken pipe
   at System.Net.Sockets.Socket.Send(ReadOnlySpan`1 buffer, SocketFlags socketFlags)
   at System.IO.Pipes.PipeStream.WriteCore(ReadOnlySpan`1 buffer)
   --- End of inner exception stack trace ---
   at System.IO.Pipes.PipeStream.WriteCore(ReadOnlySpan`1 buffer)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.IO.StreamWriter.WriteLine(String value)
   at slskd.Integrations.Scripts.ScriptService.<>c__DisplayClass10_1.<HandleEvent>b__3()
```

jpdillingham:
Closes #1278 

Documentation to follow once folks have had a chance to confirm functionality

### Linux

* [x] 'command' mode
* [x] 'args' mode
* [x] 'args list' mode
* [x] executable + args

### Windows

* [x] 'command' mode
* [x] 'args' mode
* [x] 'args list' mode
* [x] executable + args

### DIFF
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -255,13 +255,29 @@
 #         - DownloadFileComplete
 #       run:
 #         executable: cmd.exe
-#         command: '/c "echo %SLSKD_SCRIPT_DATA% >> downloads.txt"'
+#         args: '/c "echo %SLSKD_SCRIPT_DATA% >> downloads.txt"'
 #     run_with_windows_git_bash:
 #       on:
 #         - DownloadFileComplete
 #       run:
 #         executable: C:\Program Files\Git\git-bash.exe
-#         command: '-c "echo $SLSKD_SCRIPT_DATA >> downloads.txt"'
+#         args: '-c "echo $SLSKD_SCRIPT_DATA >> downloads.txt"'
+#     run_with_sh_and_args_list:
+#       on:
+#         - DownloadFileComplete
+#       run:
+#         executable: /bin/sh
+#         args_list:
+#           - -c
+#           - echo $SLSKD_SCRIPT_DATA >> downloads.txt
+#     run_with_windows_cmd_and_args_list:
+#       on:
+#         - DownloadFileComplete
+#       run:
+#         executable: cmd.exe
+#         args_list:
+#           - /c
+#           - echo %SLSKD_SCRIPT_DATA% >> downloads.txt
 #   ftp:
 #     enabled: false
 #     address: ~
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -234,17 +234,17 @@
 #       retry:
 #         attempts: 3
 #   scripts:
-#     run_command_with_linux_system_shell: # slskd uses the executable defined in $SHELL
+#     run_command_with_linux_system_shell: # automatically uses the executable defined in $SHELL
 #       on:
 #         - DownloadFileComplete
 #       run:
-#         command: 'echo "$SLSKD_SCRIPT_DATA" >> slskd_downloads.txt'
+#         command: 'echo "$SLSKD_SCRIPT_DATA" >> linux_system_shell.txt'
 #     run_with_sh:
 #       on:
 #         - DownloadFileComplete
 #       run:
 #         executable: /bin/sh
-#         args: '-c "echo $SLSKD_SCRIPT_DATA >> downloads.txt"'
+#         args: '-c "echo $SLSKD_SCRIPT_DATA >> linux_sh.txt"'
 #     run_with_sh_and_args_list:
 #       on:
 #         - DownloadFileComplete
@@ -252,8 +252,8 @@
 #         executable: /bin/sh
 #         args_list:
 #           - -c
-#           - echo $SLSKD_SCRIPT_DATA >> downloads.txt
-#     run_command_with_windows_system_shell:
+#           - echo $SLSKD_SCRIPT_DATA >> linux_sh_and_args_list.txt
+#     run_command_with_windows_system_shell: # automatically uses cmd.exe (included with all Windows installs)
 #       on:
 #         - DownloadFileComplete
 #       run:
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -238,11 +238,27 @@
 #       on:
 #         - DownloadFileComplete
 #         - DownloadDirectoryComplete
-#       run: data/my_script.sh --json-to-process $EVENT
-#     my_logging_script:
+#       run:
+#         executable: ~ # if omitted, defaults to your operating system's $SHELL, or cmd.exe on Windows
+#         command: data/my_script.sh # the script reads the data from the $SLSKD_SCRIPT_DATA envar
+#     run_with_sh:
 #       on:
-#        - All
-#       run: data/log_slskd_events.sh $DATA
+#         - DownloadFileComplete
+#       run:
+#         executable: /bin/sh
+#         command: '-c "echo $SLSKD_SCRIPT_DATA >> downloads.txt"'
+#     run_with_windows:
+#       on:
+#         - DownloadFileComplete
+#       run:
+#         executable: cmd.exe
+#         command: '/c "echo %SLSKD_SCRIPT_DATA% >> downloads.txt"'
+#     run_with_windows_git_bash:
+#       on:
+#         - DownloadFileComplete
+#       run:
+#         executable: C:\Program Files\Git\git-bash.exe
+#         command: '-c "echo $SLSKD_SCRIPT_DATA >> downloads.txt"'
 #   ftp:
 #     enabled: false
 #     address: ~
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -250,7 +250,7 @@
 #         - DownloadFileComplete
 #       run:
 #         executable: /bin/sh
-#         args_list:
+#         arglist:
 #           - -c
 #           - echo $SLSKD_SCRIPT_DATA >> linux_sh_and_args_list.txt
 #     run_command_with_windows_system_shell: # automatically uses cmd.exe (included with all Windows installs)
@@ -275,7 +275,7 @@
 #         - DownloadFileComplete
 #       run:
 #         executable: cmd.exe
-#         args_list:
+#         arglist:
 #           - /c
 #           - 'echo "%SLSKD_SCRIPT_DATA%" >> windows_cmd_and_args_list.txt'
 #   ftp:
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -238,29 +238,12 @@
 #       on:
 #         - DownloadFileComplete
 #       run:
-#         command: '"echo '$SLSKD_SCRIPT_DATA' >> slskd_downloads.txt"'
-#     run_command_with_windows_system_shell: # slskd uses cmd.exe
-#       on:
-#         - DownloadFileComplete
-#       run:
-#         command: '"echo "%SLSKD_SCRIPT_DATA%" >> slskd_downloads.txt"'
+#         command: 'echo "$SLSKD_SCRIPT_DATA" >> slskd_downloads.txt'
 #     run_with_sh:
 #       on:
 #         - DownloadFileComplete
 #       run:
 #         executable: /bin/sh
-#         args: '-c "echo $SLSKD_SCRIPT_DATA >> downloads.txt"'
-#     run_with_windows_cmd:
-#       on:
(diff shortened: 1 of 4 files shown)


config/slskd.example.yml
```diff
@@ -238,11 +238,27 @@
...
+#         executable: ~ # if omitted, defaults to your operating system's $SHELL, or cmd.exe on Windows
+#         command: data/my_script.sh # the script reads the data from the $SLSKD_SCRIPT_DATA envar
+#     run_with_sh:
 #       on:
-#        - All
-#       run: data/log_slskd_events.sh $DATA
+#         - DownloadFileComplete
+#       run:
+#         executable: /bin/sh
+#         command: '-c "echo $SLSKD_SCRIPT_DATA >> downloads.txt"'
```
boyska, line 249: this example is insecure: an attacker can create funny slskd_script_data which would allow them to execute any command on a shell (RCE).

config/slskd.example.yml
```diff
@@ -238,11 +238,27 @@
 #       on:
 #         - DownloadFileComplete
 #         - DownloadDirectoryComplete
-#       run: data/my_script.sh --json-to-process $EVENT
-#     my_logging_script:
+#       run:
+#         executable: ~ # if omitted, defaults to your operating system's $SHELL, or cmd.exe on Windows
+#         command: data/my_script.sh # the script reads the data from the $SLSKD_SCRIPT_DATA envar
```
boyska, line 243: why is this called `command`? I understand this is what is usually called `args` (except it's a single string instead of a list of strings, which is a more typical way of doing it).

boyska: and why isn't `data/my_script.sh` passed in the `executable` field? not that it would really make a difference...

src/slskd/Integrations/Scripts/ScriptService.cs
```diff
@@ -64,45 +80,40 @@ private async Task HandleEvent(Event data)
...
-                        StartInfo = new ProcessStartInfo(fileName: executable)
+                        StartInfo = new ProcessStartInfo(executable ?? DefaultExecutable)
                         {
+                            WorkingDirectory = Program.ScriptDirectory,
                             UseShellExecute = false,
+                            CreateNoWindow = true,
                             RedirectStandardOutput = true,
                             RedirectStandardError = true,
-                            RedirectStandardInput = true,
+                            Arguments = args,
```
boyska, line 109: I know nothing about csharp (so  feel free to ignore this comment), but isn't [ArgumentList](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.argumentlist?view=net-9.0#system-diagnostics-processstartinfo-argumentlist) just more practical than `Arguments`?

src/slskd/Integrations/Scripts/ScriptService.cs
```diff
@@ -64,45 +80,40 @@ private async Task HandleEvent(Event data)
...
-                    Log.Debug("Running script '{Script}': {Run}", script.Key, run);
+                    Log.Debug("Running script '{Script}': \"{Executable}\" {Args} (id: {ProcessId})", script.Key, executable, args, processId);
                     var sw = Stopwatch.StartNew();
 
                     process = new Process()
                     {
-                        StartInfo = new ProcessStartInfo(fileName: executable)
+                        StartInfo = new ProcessStartInfo(executable ?? DefaultExecutable)
                         {
+                            WorkingDirectory = Program.ScriptDirectory,
```
boyska, line 104: What's the rationale for having a different WorkingDirectory?
I don't see any problem with it, but I also don't see any advantage

src/slskd/Integrations/Scripts/ScriptService.cs
```diff
@@ -64,45 +80,40 @@ private async Task HandleEvent(Event data)
...
                             UseShellExecute = false,
+                            CreateNoWindow = true,
                             RedirectStandardOutput = true,
                             RedirectStandardError = true,
-                            RedirectStandardInput = true,
+                            Arguments = args,
                         },
                     };
 
+                    process.StartInfo.Environment.Clear();
```
SevereOverfl0w, line 113: Clearing the environment completely might cause problems. On Linux it tells you things like who the current user is, where to find executables, etc.

config/slskd.example.yml
jpdillingham: thanks, missed that

src/slskd/Integrations/Scripts/ScriptService.cs
jpdillingham: it would be but unfortunately there's only `Arguments` which is a string

boyska: What about that? https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.-ctor?view=net-9.0#system-diagnostics-processstartinfo-ctor(system-string-system-collections-generic-ienumerable((system-string))) Again, sorry if I'm talking gibberish: I really know nothing about the .net environment

config/slskd.example.yml
boyska: I'm not sure this is the best idea, but if you really think you should, then you should also go into the string quoting headache.

Let's assume you want to do something very simple, like this

```sh
echo "The new value is $(echo -n "$SLSKD_SCRIPT_DATA" | jq '.foo[]|.bar')" >> /tmp/slskd-log.txt
```

This needs two different quoting: one is done by the user, when writing the YAML configuration. It sucks, but it's not your problem as a developer.

The second kind of quoting happens when you want to invoke such a line. I see you can't have ArgumentList, so you're left with a single string. So you would need to do some form of quoting so that ProcessStartInfo splits your command the correct way. (ie: `["/bin/sh", "-c", "whatever I wrote earlier"]`).  I don't know how that works exactly, and I wouldn't be surprised if that was platform-specific.

src/slskd/Integrations/Scripts/ScriptService.cs
jpdillingham: creating any files from the command drops them in the executable's working directory, which is likely inside of a container, potentially with a read only filesystem, or maybe in a read only location on disk.  doing it this way will drop them into the user's /app directory, which is expected to be writable and is validated at startup.

mostly a 'pit of success' for less technical users

src/slskd/Integrations/Scripts/ScriptService.cs
jpdillingham: i was thinking it would enhance security since the script would only have access to the data SLSKD set; sort of a 'least privileged' sort of thing.  in reality the shell process implicitly loads all environment variables on both Windows and Linux and this doesn't really do much unless someone is using an executable other than a shell.  i pulled it out for now.

config/slskd.example.yml
```diff
@@ -255,13 +255,29 @@
...
+#           - -c
+#           - echo $SLSKD_SCRIPT_DATA >> downloads.txt
+#     run_with_windows_cmd_and_args_list:
+#       on:
+#         - DownloadFileComplete
+#       run:
+#         executable: cmd.exe
+#         args_list:
+#           - /c
+#           - echo %SLSKD_SCRIPT_DATA% >> downloads.txt
```
boyska, line 280: that's sooo much nicer interface!

is there any case in which using a single string (`args`) should be recommended for user? shouldn't `args_list` just be enough for everyone, allowing for the code to be simplified a lot?

src/slskd/Integrations/Scripts/ScriptService.cs
```diff
@@ -115,7 +115,7 @@ private async Task HandleEvent(Event data)
                                 CreateNoWindow = true,
                                 RedirectStandardOutput = true,
                                 RedirectStandardError = true,
-                                Arguments = run.Command.StartsWith(DefaultCommandPrefix) ? run.Command : $"{DefaultCommandPrefix} {run.Command}",
+                                Arguments = run.Command.StartsWith(DefaultCommandPrefix) ? run.Command : $"{DefaultCommandPrefix} \"{run.Command}\"",
```
boyska, line 118: Wouldn't this break if the user supplied a command which includes double quotes? ie:

```sh
touch "/tmp/notify me"
```

jpdillingham: it didn't when i was testing on Linux last night.  i'm going to test on Windows this morning

config/slskd.example.yml
jpdillingham: My big concern with this is that if `args` were to take on the behavior of `args_list` (which I've just renamed to `arglist` to better match other applications) people would fumble trying to specify a string where an array is expected.  YAML formatting errors get really ugly and cryptic error messages and I see users struggle with it a lot, so I'm trying my best to make it as intuitive as possible

src/slskd/Integrations/Scripts/ScriptService.cs
jpdillingham: Broke on Windows; I pushed a fix.  Thanks for pointing it out.

jpdillingham:
I'm going to push this out to the canary so I can test in a wider variety of scenarios.  The final structure is:

 * 'Command mode': specify only a single string that will be passed as an argument to your system shell prefixed by `-c` or `/c` depending on your OS.  This should work more or less like a command line, and users are on the hook to get quotes right.
 * 'Args mode': specify an executable (or not) and a single string that will be passed as an argument.  I'm anticipating that this would be used with a third party executable (not a shell) and users have full control over the args.  This is included as a 'backstop' to help users that may rely on their intuition (rather than the docs) to help them configure scripts.
 * 'Args list mode': specify an executable (or not) and an array of arguments.  Users _are not_ on the hook for getting quotes right.  This is the recommended way to configure a script with non-trivial args.

I'm going to hold off on updating the docs until I've verified that things are still working as expected.

jpdillingham:
Thanks @SevereOverfl0w and @boyska for the feedback!  Please let me know what you think of the finished product.

config/slskd.example.yml
boyska: > people would fumble trying to specify a string where an array is expected

I see your point. My experience, as a user, is that I learned why "strings as command arguments don't work" the hard way: I was trying to write commands in similar settings, and that was just repeatedly buggy. And whenever I was asking developers what I should do to handle the n-th level of quoting, the only meaningful tip I got was "just create a wrapper script and invoke that". So, since then, my thought is always: why isn't this just stated upfront since the beginning? Why are examples pointing me in a direction which is fragile and sometimes just not working? Why do people mess with placeholder in argument strings (which are difficult to handle) compared to passing data to stdin (which is straightforward)?

Anyway, back to the case at hand, I see how your concern is an important one, too. I still stand that it would be better to just have `args_list`. I think you are right in pointing out that users could get confused. Good mitigations are:
 - provide clear examples
 - provide a clear error message, like "args_list should be a list of arguments, not a string. See link.to/doc/ for examples"

I think you are still headed to have this [triple mode](https://github.com/slskd/slskd/pull/1292#issuecomment-2692497083) merged. I think it's making the code more complex for no real value, but you're the maintainer, so it's your call :)

PR #1298, opened 2025-02-23
Adding context menu to rooms with reply, view user profile, and browse shares

samhatter:
# Chat Room Context Menus
## Description
This PR adds a context menu for messages in chat rooms. The context menu adds the ability to format a reply, view the user's profile, or browse their shares. The context menu can be accessed by right clicking on a chat message in a room.
## Justification
Modern UI features are desired by many soulseek users who are hesitant to adopt slskd even if they understand the architectural advantages (see prevalence of hacky nicotine clients on dockerhub). A context menu to make the chat feel more user-friendly could attract users and I have had fun using it thus far.
## Considerations
I am trying to make a feature to ignore users, but it will be a more significant challenge to implement the backend endpoint to add a user to a blacklist. Current solutions for both user and managed blacklists seem improper for this task.
## Technical Information
In the Room component, I have added onContextMenu handlers on room messages that render a Portal (semantic-ui)  at the coordinates of the click. This menu contains buttons that can navigate the UI with react-router-dom and pass the necessary state. Browse and Users components are updated to receive the state and display it properly.

@jpdillingham

### DIFF
--- a/src/web/src/components/Browse/Browse.jsx
+++ b/src/web/src/components/Browse/Browse.jsx
@@ -6,6 +6,7 @@ import Directory from './Directory';
 import DirectoryTree from './DirectoryTree';
 import * as lzString from 'lz-string';
 import React, { Component } from 'react';
+import { withRouter } from 'react-router-dom';
 import { Card, Icon, Input, Loader, Segment } from 'semantic-ui-react';
 
 const initialState = {
@@ -42,6 +43,9 @@ class Browse extends Component {
       },
       () => this.saveState(),
     );
+    if (this.props.location.state?.user) {
+      this.setState({ username: this.props.location.state.user }, this.browse);
+    }
 
     document.addEventListener('keyup', this.keyUp, false);
   }
@@ -150,11 +154,13 @@ class Browse extends Component {
 
   loadState = () => {
     this.setState(
-      JSON.parse(
-        lzString.decompress(
-          localStorage.getItem('soulseek-example-browse-state') || '',
-        ),
-      ) || initialState,
+      (!this.props.location.state?.user &&
+        JSON.parse(
+          lzString.decompress(
+            localStorage.getItem('soulseek-example-browse-state') || '',
+          ),
+        )) ||
+        initialState,
     );
   };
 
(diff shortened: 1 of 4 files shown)


davidarkemp:
Hey @samhatter this all looks really good. 
I have a few questions - they are really the fine points, as this could be merged now and most users would benefit.

- Could the `renderContextMenu` be extracted to a separate component to prevent it from re-rendering when other state in the `Rooms` component changes? I realise this is a really nit-picky, but it's the kind of thing that comes back to bite you with React. 
- I like how you've implemented the context menu, but did you consider using [`floating-ui`](https://floating-ui.com/)? It handles a lot of edge-cases
- How do touch-screen users get the context menu?
- If you browse a user from the rooms context and it's the same user as you've just been browsing, you loose all the state. Was this intended?

samhatter:
All very good questions!

> * Could the `renderContextMenu` be extracted to a separate component to prevent it from re-rendering when other state in the `Rooms` component changes? I realise this is a really nit-picky, but it's the kind of thing that comes back to bite you with React.

I believe the rending logic could indeed be put into its own components. I didn't do that because that seems to be the only part of this implementation that is portable, but I do see the utility in making it easier to implement context menus in other components.

> * I like how you've implemented the context menu, but did you consider using [`floating-ui`](https://floating-ui.com/)? It handles a lot of edge-cases

I did not consider any new ui libraries since I was unsure of your stance on adding additional ui libraries. For consistency, I decided to build it with semantic-ui. However, I have taken into consideration a few edge cases (menu is scrollable if it overflows, works on mobile)

> * How do touch-screen users get the context menu?

Since the context menu is triggered by an onContextMenu event, touch screen users should be able to access context menus intuitively by clicking and holding on the message elements, the same as any  context menu for them. In truth I have only tested this in my devtools, but I could host my branch and test natively on my iPhone.

> * If you browse a user from the rooms context and it's the same user as you've just been browsing, you loose all the state. Was this intended?

This is intentional, good catch. I did this so that if i navigated from a context menu it wouldn't try and load the state, displaying old info before overriding that and navigating away to a new browse request. However, now that you mention it, I could probably add an additional check to add a case where the saved state is for the same user as the context menu browse request. That shouldn't be too difficult to get right.

jpdillingham:
Thanks for sending this in @samhatter, I'll take a look this weekend!

Issue #1291, opened 2025-02-13
OOM when launching (0.22.2)


mrnoname1000:
I've left slskd running with Docker for several months without checking on it and now it seems to have entered an endless crashloop. Here are some logs:

<details>
<summary>Logs (DEBUG=true)</summary>

```
at slskd.Program.Main(String[] args)


                    ▄▄▄▄     ▄▄▄▄     ▄▄▄▄
              ▄▄▄▄▄▄█  █▄▄▄▄▄█  █▄▄▄▄▄█  █
              █__ --█  █__ --█    ◄█  -  █
              █▄▄▄▄▄█▄▄█▄▄▄▄▄█▄▄█▄▄█▄▄▄▄▄█
╒════════════════════════════════════════════════════════╕
│           GNU AFFERO GENERAL PUBLIC LICENSE            │
│                   https://slskd.org                    │
│                                                        │
│              0.22.2.0 (0.22.2.0+a2e1e65b)              │
└────────────────────────────────────────────────────────┘
[12:04:44 INF] Version: 0.22.2.0 (0.22.2.0+a2e1e65b)
[12:04:44 INF] System: .NET 8.0.12, Unix 6.9.7.64, 64 bit, 40 processors
[12:04:44 INF] Process ID: 7 (64 bit)
[12:04:44 INF] Invocation ID: a1da1c39-fa36-4b1c-9c60-2a0232ac0388
[12:04:44 INF] Instance Name: default
[12:04:44 INF] Configuring application...
[12:04:44 INF] Using application directory /app
[12:04:44 INF] Using configuration file /app/slskd.yml
[12:04:44 INF] Storing application data in /app/data
[12:04:46 INF] Listening for HTTP requests at http://0.0.0.0:5030/
[12:04:46 INF] Using base url /
[12:04:46 INF] Serving static content from /slskd/wwwroot
[12:04:46 INF] Configuration complete.  Starting application...
[12:04:46 INF] Blacklist disabled
[12:04:46 INF] Application started
[12:04:57 ERR] An exception occurred while iterating over the results of a query for context type 'slskd.Transfers.TransfersDbContext'.
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at Microsoft.Data.Sqlite.SqliteValueReader.GetGuid(Int32 ordinal)
   at lambda_method19(Closure, QueryContext, DbDataReader, ResultContext, SingleQueryResultCoordinator)
[...]

Issue #1280, opened 2025-01-21
Entity Framework, EnumToStringConverter, HasFlag(), and you


jpdillingham:
I've referenced EF issue https://github.com/dotnet/efcore/issues/10434 a few times in application logic and every time I see it I go down a 10 minute rabbit hole trying to decipher what it means, so I'm creating this issue so I can swap the cryptic reference for this one.

Some DbContexts are configured to store enums as strings, specifically `TransfersDbContext`, but potentially others (not `SearchDbContext`, for the time being).  This causes numeric enum values to be stored in the database as their string representations, which are more friendly to human reviewers or (less than ideally) external applications interested in direct database interop.

The problem with this approach, and the reason EF issue 10434 is linked, is that `HasFlag` doesn't work on these fields because of the string conversion; EF converts the `HasFlag` predicate into a bitwise expression expecting an integer, and it finds strings instead.  This doesn't cause any problems other than the predicate not actually being applied to the query, resulting in an over-fetch of data and unexpected results.

To get around this I'm choosing to include the `HasFlag` predicate in the query so that if this capability is added in the future the over-fetch of data will cease without me needing to do anything.  To ensure the application works as expected, I'm applying another `Where` to the result that filters the data a second time.

So, in summary:

* When:
  * Writing an EF expression containing `HasFlag`
  * Against a `DbContext` using an `EnumToStringConverter`
* Then:
  * Include `HasFlag` in the EF expression
  * Filter the data after the fetch using the same `HasFlag` expression

If/when EF implements this capability, we can stop including the redundant `Where()`.

jpdillingham:
As I'm reading what I wrote, it seems like I should probably remove `EnumToStringConverter` to avoid this footgun altogether.  This will make database interop more difficult but I think that's a reasonable tradeoff.

If I were to attempt this, I'd have to figure out database migrations; something I've successfully avoided so far.

davidarkemp:
@jpdillingham IMHO, databases and Enum Flags really don't work well together. And other languages (like JavaScript) don't have the concept of Enums, nor flags. TS fudges Enums.

I'm trying to work out what the different states are, but from looking at the SoulSeek code client, I can see:

- Queued Locally
- Queued Remotely
- Requested
- Initializing
- In Progress
- Completed
  - Succeeded
  - Timed Out
  - Cancelled
  - Rejected
  - Aborted
  - Errored 

I tried to draw a state diagram of a download:

```mermaid
stateDiagram-v2
[...]

PR #1343, opened 2025-05-03
Add a system to perform database migrations

jpdillingham:
A couple of unfortunate decisions have led me to need to change how the `Transfers` database is laid out, specifically due to #1280 (database/C# enums), which is symptomatic in #1291 (OOM issue with Transfers).

I have been putting this off, hoping that it wouldn't be needed, but the time has come.

This will be a "major" version update and might cause significant headaches for folks on low-spec hardware and a lot of historical transfer data.

## Testing

* [x] Successfully migrates an existing database (with ~250k records)

```
[17:38:16 WRN] 1 outstanding database migration(s) to apply. This operation must be completed before the application can start.
[17:38:16 WRN] --> The ID for this migration is 052525_103816. Use it to locate pre-migration database backups, should manual cleanup be needed. <--
[17:38:16 INF] Backing up existing databases...
[17:38:16 INF] Backed database C:\Users\JP\AppData\Local\slskd\data\search.db up to C:\Users\JP\AppData\Local\slskd\data\backups\search.pre-migration-backup.052525_103816.db
[17:38:16 INF] Backed database C:\Users\JP\AppData\Local\slskd\data\transfers.db up to C:\Users\JP\AppData\Local\slskd\data\backups\transfers.pre-migration-backup.052525_103816.db
[17:38:16 INF] Backed database C:\Users\JP\AppData\Local\slskd\data\messaging.db up to C:\Users\JP\AppData\Local\slskd\data\backups\messaging.pre-migration-backup.052525_103816.db
[17:38:16 INF] Backed database C:\Users\JP\AppData\Local\slskd\data\events.db up to C:\Users\JP\AppData\Local\slskd\data\backups\events.pre-migration-backup.052525_103816.db
[17:38:16 INF] Databases backed up successfully
[17:38:16 WRN] Beginning migration(s)
[17:38:16 WRN] This may take some time. Avoid stopping the application before the process is complete. If the process is interrupted it may be necessary to manually revert to backups.
[17:38:16 WRN] Applying migration Z04012025_TransferStateMigration (1 of 1)
[17:38:16 INF] > Schema changes applied, copying data...
[17:38:19 INF] > Data copied, adjusting columns...
[17:38:19 INF] > Updating column 1 of 1. This may take some time...
[17:38:20 INF] > Done!
[17:38:20 INF] Migration Z04012025_TransferStateMigration was applied successfully (elapsed: 4346ms)
[17:38:20 INF] 1 migration(s) applied successfully (elapsed: 4347ms)
[17:38:20 INF] Migration(s) complete!
```

* [x] Backs up (and leaves) databases

image

* [x] Migrated database has the right data layout (State is now a number, StateDescription contains the text)

image

image

* [x] Newly downloaded/uploaded records have the right layout
* [x] App creates the correct file if one doesn't exist at startup (starting from scratch)


src/slskd/Core/Data/Migrations/Z04012025_TransferStateMigration.cs
```diff
@@ -0,0 +1,208 @@
...
+            using var copyCommand = new SqliteCommand(copy, connection, transaction);
+            copyCommand.ExecuteNonQuery();
+
+            Log.Information("Data copied, adjusting columns...");
+
+            /*
+                we no longer need the old table, so drop it
+            */
+            using var dropCommand = new SqliteCommand("DROP TABLE Transfers_old;", connection, transaction);
+            dropCommand.ExecuteNonQuery();
```
wburklund, line 165: suggestion: Add a column to Transfers instead of replacing the whole table.

note: SQLite's support of transactional DDL limits the corner cases that could crop up

src/slskd/Core/Data/Migrator.cs
```diff
@@ -0,0 +1,276 @@
...
+    {
+        Dictionary<string, DateTime> history = [];
+
+        Log.Information("Checking for outstanding database migrations...");
+
+        /*
+            load migration history from the history file in the root of the data directory. this file contains a key/value
+            pair for each migration that's been applied, along with the migration date.
+            if force=true, we're going to disregard history and overwrite it.
+        */
```
wburklund, line 88: suggestion: Store migration history in a database table like most migration systems (EF, Alembic, Liquibase, ...)

src/slskd/Core/Data/Migrations/Z04012025_TransferStateMigration.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        while (reader.Read())
+        {
+            var state = reader.GetString(0);
+            dict[state] = (int)Enum.Parse(typeof(TransferStates), state);
+        }
+
+        Log.Debug("State -> int map: {Map}", dict);
+        return dict;
+    }
```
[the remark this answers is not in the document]
jpdillingham: great point.  i was cringing at needing to generate a list of all possible enum combinations but it should actually be pretty straightforward since i can just increment an integer and loop

src/slskd/Core/Data/Migrations/Z04012025_TransferStateMigration.cs
jpdillingham: the new table is so that the `State` and `StateDescription` columns will be next to each other.  i'm thinking maybe it's not worth the extra effort (and file i/o)

src/slskd/Core/Data/Migrations/Z04012025_TransferStateMigration.cs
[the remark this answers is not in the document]
jpdillingham: there are 8191 possible combinations, and only a couple of dozen make sense.  i'm afraid i'll miss one and break somebody, so what i've done instead is copy/paste the enum into the migration so it will remain stable.

appreciate you pointing this out!

src/slskd/Core/Data/Migrator.cs
jpdillingham: i think i will; originally i didn't because Entity Framework would make it a hassle, but i don't have to use EF for this

aaronjwood: Since this topic came up, would it save you a lot of (ongoing) effort to use something like Flyway, or the above mentioned? I've used Flyway in a commercial setting and it worked fine, though it was heavy. I've heard https://github.com/golang-migrate/migrate is much better but I can't say I've tried it myself.

These sorts of tools are nice because you can usually stay away from your codebase and keep the separation of concerns very clean.

PR #1383, opened 2025-07-04
Fix races

aaronjwood:
These had protections in other areas but were missing the same guards as part of the getters/setters.


jpdillingham:
Thanks for the PR but It's not necessary to synchronize properties; setters are simple assignments under the hood and there's no risk of a race condition

PR #1411, opened 2025-08-04
Fix bug preventing proper change detection when updating options with an enum type

jpdillingham:
I noticed this morning that changing the Soulseek diagnostic level from `Info` to `Debug` didn't update the UI or tell me I had to restart.  Digging further I found that the 'diff' that is done between the old and new configuration didn't detect the change, and the underlying cause is that the diff logic ignores enum values (rather, tries to traverse them).

`DiagnosticLevel` is the only option that's backed by an enum, so it is the only one affected.


jpdillingham:
It's now a non-issue; I update this property to use a string as a backing type and it's now consistent with the other enum based settings.

Issue #887, opened 2023-03-09
Suggestion: ability to set default search result sort


neilbaldwin:
I find that 99% of the time I'd rather see the search results ordered by queue length rather than downlaod speed but ordered by download speed is the default.

Would be good to have this in the config somewhere so you could set it to your preference.

jpdillingham:
Thanks for the suggestion.  I've been meaning to figure out how to incorporate UI preferences for a while.  I always intended to make the app work with multiple users (from the UI perspective, using the same Soulseek account under the hood), and preferences ties in with this so I have been putting it off.

jpdillingham:
I created #888 to capture the requirements for multiple users.  As I'm thinking through that, I'll figure out how to store and manage user "preferences" (as opposed to application options/config)

neilbaldwin:
Ah, superb, thank you!

jpdillingham:
Let's keep this open so I don't forget!

neilbaldwin:
Oops, sorry! I was a bit over enthusiastic there :)

PR #1412, opened 2025-08-04
Default web UI search filter

### LINKED ISSUE #813: Add capability to save search filter as default
I think it would be very helpful to allow users to save an entered search filter as default.

For example, if I only ever care about lossless files, I could enter "islossless" in the filter and save this as the default, and then subsequent searches would have this pre-filled and only show lossless files.

hc-nolan:
Closes #813

Introduces a new config option that gets passed to the React app and used as the initial value for `resultFilters`. Let me know if there are any changes you would like!

### DIFF
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -146,6 +146,7 @@
 #   url_base: /
 #   content_path: wwwroot
 #   logging: false
+#   search_filter:
 #   authentication:
 #     disabled: false
 #     username: slskd
--- a/src/slskd/Application.cs
+++ b/src/slskd/Application.cs
@@ -1466,6 +1466,7 @@ namespace slskd
 
                 Metrics.Search.ResponseLatency.Observe(sw.ElapsedMilliseconds);
                 Metrics.Search.CurrentResponseLatency.Update(sw.ElapsedMilliseconds);
+                Metrics.Search.RequestCount.CountUp();
 
                 if (results.Any())
                 {
--- a/src/slskd/Application.cs
+++ b/src/slskd/Application.cs
@@ -1433,6 +1433,7 @@ namespace slskd
         private async Task<SearchResponse> SearchResponseResolver(string username, int token, SearchQuery query)
         {
             Metrics.Search.RequestsReceived.Inc(1);
+            Metrics.Search.CurrentRequestRate.CountUp(1);
 
             if (Users.IsBlacklisted(username))
             {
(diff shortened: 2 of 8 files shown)


hc-nolan:
Just realized I forgot to update the docs; will do that shortly.

hc-nolan:
After reviewing other issues, I could add a 2nd new option that handles #887 as well. Implementation would essentially be the exact same; new config option in web options that can be either uploadSpeed or queueLength, that gets passed to the definition of `resultSort` as the initial state. 

If this sounds good to you just let me know; I can either open a new PR or add it to this one.

jpdillingham:
Thanks for putting this up!  I _think_ I like the approach using HTML injection; I hadn't considered that before.

I will pull this locally and play around a bit.  My only concerns:

* The user experience when changing the value (I think a refresh is probably required) and when someone wants do override or remove the default
* Organizing the option(s) to account for organic growth (breaking changes if we get it wrong or change our minds)

jpdillingham:
An alternative to using HTML injection would be to stick frontend preferences into the state, or wire up an additional route and/or SignalR hub so the UI can fetch them when it needs them.

I had been putting off thinking about this because I had always imagined slskd to be a multi-user application (e.g. allow a spouse to search and download things, but not have full control) and storing user preferences in the backend makes that difficult.  That's not a great reason to not do something in the short term, though.

hc-nolan:
The only reason I went with HTML injection is because it was what was already there for the baseUrl and port, but yeah, I definitely didn't consider that this could be a per-user config. I would totally be willing to help implement a user settings system to facilitate this and other settings for the future though.

I am thinking about how to handle it, so going through the code that handles authentication - am I correct that right now it only supports the single user? 

If so, I suppose the path forward depends on whether you would like to proceed with facilitating multiple users, and if so how secure you want to handle it. Apart from storing user settings like this in the frontend state, the simplest backend solution I can think of is just a separate per-user YAML config containing these settings, which can be loaded after the user logs into the web UI. This sort of approach would also facilitate backend per-user settings in the future.

We could slightly modify the current username and password config options to be comma delimited and change the authentication code slightly to split the values by comma and compare the user-provided login information to everything in that list 1-by-1, but this is obviously not secure at all and is the absolute laziest solution :) 

I'd be down to put more proper thought into how to facilitate a secure user authentication and configuration scheme too, but if so then we should probably take this conversation to a separate issue as it will require much more discussion :laughing:

PR #1413, opened 2025-08-04
Substantially improve performance of search request filtering

jpdillingham:
Long story short, I discovered the CPU utilization rose substantially for each entry in the list of incoming search request filters, as did garbage collection pressure.  I finally tracked it down to inefficiencies of `IEnumerable` handling; to iterate over one (such as `Any()` in this example) a whole state machine is created, and this can't be optimized out because of the dynamic nature.

I switched the type of collection for the compiled filter regular expressions from `IEnumerable` to `Regex[]` and the computation time for filtering dropped from ~45ms for a list of 30 filters to under 1ms (not sure how many ns but negligible).

image

Here we see, from left to right:

* 2 search filters
* Distributed network disabled (trying to find the source)
* 15 search filters
* 0 search filters

I'm positive that I've identified the culprit and have verified locally the benefits to this change.  I'm expecting performance with just about any number of filters to be the same as with zero filters once this change is deployed.

### DIFF
--- a/src/slskd/Application.cs
+++ b/src/slskd/Application.cs
@@ -135,7 +135,7 @@ namespace slskd
                 regexOptions |= RegexOptions.IgnoreCase;
             }
 
-            CompiledSearchResponseFilters = OptionsAtStartup.Filters.Search.Request.Select(f => new Regex(f, regexOptions));
+            CompiledSearchRequestFilters = OptionsAtStartup.Filters.Search.Request.Select(f => new Regex(f, regexOptions)).ToArray();
 
             State = state;
             State.OnChange(state => State_OnChange(state));
@@ -228,7 +228,7 @@ namespace slskd
         private ISearchService Search { get; set; }
         private IRelayService Relay { get; set; }
         private IMemoryCache Cache { get; set; } = new MemoryCache(new MemoryCacheOptions());
-        private IEnumerable<Regex> CompiledSearchResponseFilters { get; set; }
+        private Regex[] CompiledSearchRequestFilters { get; set; }
         private IEnumerable<Guid> ActiveDownloadIdsAtPreviousShutdown { get; set; } = Enumerable.Empty<Guid>();
         private Options.FlagsOptions Flags { get; set; }
         private IReadOnlyCollection<string> ExcludedSearchPhrases { get; set; } = Enumerable.Empty<string>().ToList();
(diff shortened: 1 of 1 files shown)


2048khz-gachi-rmx:
I'm pretty sure the real performance culprit comes from compiling new Regex instances over and over on each iteration, not from IEnumerable inefficiencies (i assume you meant that?). Compiled Regex'es are (usually) faster than interpreted, but take a while to actually compile, so instantiations should be done once (like on startup).

```cs
using System;
using System.Linq;

class TestClass
{
    public TestClass(int idx)
        => Console.WriteLine($"Instantiated new class: {idx}");
}

class Program
{
    static void Main(string[] args)
    {
        var range = Enumerable.Range(0, 10).Select(x => new TestClass(x));
        range.Any(_ => false);
        range.Any(_ => false);
        range.Any(_ => false);
    }
}
```

Good job on spotting the issue though; i was beginning to have concerns regarding slskd's CPU usage with barely any activity

jpdillingham:
These have always been compiled once when the app starts (or when options change): https://github.com/slskd/slskd/blob/2bb92971e5e6414ad1d01aaca92808cbfe96fc72/src/slskd/Application.cs#L131-L138

The only real difference was the `.ToArray()`.  I think there may be a nuanced issue with using `IEnumerable<Regex>` for the property type but I didn't test changing just that.

Garbage collection pressure increased along with CPU, which makes me think there's some sort of synchronous collection going on along with the state machine lifecycle under the hood.

2048khz-gachi-rmx:
> The only real difference was the .ToArray().

IEnumerable produced by LINQ has [deferred evaluation](https://learn.microsoft.com/en-us/dotnet/standard/linq/deferred-execution-lazy-evaluation); it won't actually execute the expression until something forces it to materialize (`.Any()`, `.ToList()`, etc.). So, because `CompiledSearchResponseFilters` was storing an unmaterialized IEnumerable, each time something iterated over it (like calling `.Any()`), it was creating its' contents anew. And each time it did that, it was applying the `.Select(f => new Regex(f, regexOptions))`, thus recompiling all the regexes, which is slow

Just for demonstration purposes, you can try undoing the changes and then _also_ get rid of the `.Compiled` flag. It should be faster, since it's not recompiling Regex anymore.

jpdillingham:
That's right, I forgot about that.  I was just looking for similar optimizations in Soulseek.NET and was seeing `.ToList()` everywhere inside of constructors, so I knew this at one point :)

PR #1420, opened 2025-08-14
fix(docs): add missing prefixes to some environment variables

RalphORama:
Some of the environment variables in the configuration doc were missing the `SLSKD_` prefix, i.e. `ROOMS` should be `SLSKD_ROOMS`. I encountered this while setting up slskd with Docker Compose, as ROOMS were not auto-joining. I tested all of these changes and they all require the `SLSKD_` prefix in order to function correctly.

I also cleaned up the alignment in a few markdown tables while i was editing the file.


jpdillingham (approved):
Thanks!

PR #1434, opened 2025-08-31
#813 allow saving and loading filters in local storage

antonmihaylov:
Aims to solve https://github.com/slskd/slskd/issues/813 by introducing a way to save and load filters from the browser's local storage. 

Demo:
https://github.com/user-attachments/assets/3f4b10f2-2c65-4db0-896b-7fe7fc6643e5

### DIFF
--- a/src/web/src/config.js
+++ b/src/web/src/config.js
@@ -5,8 +5,18 @@ const rootUrl =
   process.env.NODE_ENV === 'production'
     ? urlBase
     : `http://localhost:${developmentPort}${urlBase}`;
-const apiBaseUrl = `${rootUrl}/api/v0`;
-const hubBaseUrl = `${rootUrl}/hub`;
+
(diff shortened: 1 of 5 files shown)


src/web/src/config.js
```diff
@@ -5,8 +5,18 @@ const rootUrl =
   process.env.NODE_ENV === 'production'
     ? urlBase
     : `http://localhost:${developmentPort}${urlBase}`;
-const apiBaseUrl = `${rootUrl}/api/v0`;
-const hubBaseUrl = `${rootUrl}/hub`;
+
+const apiBaseUrlFromLocalStorage = localStorage.getItem('apiBaseUrl');
```
antonmihaylov, line 9: not related to the issue, but i found it easier to be able to override the API url so i don't have to run the entire backend locally just for a frontend change

PR #1453, opened 2025-09-21
Support the ability for the web interface to listen on a unix socket

### LINKED ISSUE #1452: Ability to listen on unix sockets for web interface
Currently, I do not believe that it is possible for the slskd web interface to listen on a unix socket, e.g. `/run/slskd/slskd.sock`. This feature would be beneficial for setups where opening a network port is un-ideal (such as running behind network namespaces), and can allow for more secure communication between network services running on a machine.

Personally, I'd like for slskd's web interface to listen on a socket to then be reverse proxied by Caddy. This would significantly reduce the complexity of my setup, which utilizes some messy workarounds to achieve this through network namespaces. Rather than specifying ports that are open to the default namespace, then accessing the namespace address associated with slskd, I could simply reverse proxy to this socket from any namespace without the need to open any ports.

Other users can use sockets to avoid opening more network ports, or to specify which users/groups can access the service. There are arguments that this can result in better proxy performance due to the lack of networking overhead, however this can be debated.

Here's an example of what that'd look like through Caddy:
```caddy
reverse_proxy unix//run/slskd/slskd.sock {
    header_up Upgrade "websocket"
    header_up Connection "Upgrade"
}
```

encode42:
Closes #1452

I decided to work on this myself, as it was quite easy to implement. This hasn't been thoroughly tested, however it's currently running on my machine as described in the original issue!

### DIFF
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -143,6 +143,7 @@
 #     certificate:
 #       pfx: ~
 #       password: ~
+#   socket: ~
 #   url_base: /
 #   content_path: wwwroot
 #   logging: false
--- a/docs/config.md
+++ b/docs/config.md
@@ -802,6 +802,7 @@ Logging of HTTP requests is disabled by default.
 | Command-Line      | Environment Variable       | Description                                       |
 | ----------------- | -------------------------- | ------------------------------------------------- |
 | `-l\|--http-port` | `SLSKD_HTTP_PORT`          | The HTTP listen port                              |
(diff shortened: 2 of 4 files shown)


jpdillingham:
Looks good, thanks!  I have just a couple nitpicks about placement (would like socket and port related stuff side by side), but I'll fix that after.

encode42:
Since the socket is unaffected by the port/HTTPS options, I figured to place it after each, since the port does impact HTTPS. `port -> https | socket` 😁

PR #1458, opened 2025-09-28
Refine support for unix domain sockets (UDS)

jpdillingham:
Expanding on #1453, this PR just moves some things around and adds validation for the option.


jpdillingham:
FYI @encode42

PR #1512, opened 2025-11-06
Feature: Button to clear Searches

### LINKED ISSUE #1315: Button to clear Searches
I'm using slskd very happily!

I've connected it to several script, that does their own searches in a scheduled ways.

However, this clog too much for my tastes the main Search screen:

Image

And seems that there's no way to delete them manually/automatically all at once

In "System/Data/Transfer Data" there are buttons to clear the Download and Upload lists, but there's nothing about Search.

Is possible to insert something that help me clearing these?

Thanks <3

kdssoftware:
Closes #1315 

it adds a buttons to the search list page to clear all searches

test plan:
( screen recording)
https://github.com/user-attachments/assets/5ccf4ff1-642f-4592-97c3-4d3bc4f3a48f

Screenshot 2025-11-06 at 10 41 13

Screenshot 2025-11-06 at 10 41 20

### DIFF
--- a/src/web/src/components/Search/Searches.jsx
+++ b/src/web/src/components/Search/Searches.jsx
@@ -248,6 +248,19 @@ const Searches = ({ server } = {}) => {
           ref={inputRef}
           size="big"
         />
+
+        <Button
+          className="px-2"
+          disabled={!Object.keys(searches).length}
+          onClick={() => {
+            for (const search of Object.values(searches)) {
+              remove({ id: search.id });
+            }
+          }}
+          type="button"
+        >
+          Clear searches
+        </Button>
       </Segment>
       {Object.keys(searches).length === 0 ? (
         <PlaceholderSegment


src/web/src/components/Search/Searches.jsx
```diff
@@ -248,6 +248,19 @@ const Searches = ({ server } = {}) => {
           ref={inputRef}
           size="big"
         />
+
+        <Button
+          className="px-2"
+          disabled={!Object.keys(searches).length}
+          onClick={() => {
+            for (const search of Object.values(searches)) {
```
jpdillingham: i think this logic is probably fine, but would prefer to see it extracted to a `removeAll` function

jpdillingham:
I think the button is too prominent for how often it would likely be clicked.  Instead I think a drop-down menu would be better:

image

This is an example from the Semantic UI docs, about halfway down this page: https://semantic-ui.com/modules/dropdown.html

A button using the `ellipsis vertical` icon where you've got 'Clear searches' right now would be good, with just one menu item (for now): 'Delete All'

Mateleo:
How is this issue going ?

jpdillingham:
It's not a priority.  If you're struggling with this I suggest you set the search retention setting so they are deleted automatically when they expire.

cryobry:
Why not add multi-select boxes (including a select all header box) and re-search/remove/etc buttons, to mimic the downloads page and also allow users to clear multiple searches instead of all?

PR #1516, opened 2025-11-08
Add context menu for ignoring users

samhatter:
# Add Context Menu for Ignoring Users

## Description

Adds an option in the context menu to ignore users. Feature works by modifying blacklisted user group in config.

## Changes

- Adds ignore feature to context menu
- Refactors context menu and adds it to dm chat
- Better logic for positioning context menu

## Justification

I had a hard time chosing how to implement the ignore feature since slskd has two methods of ignoring a user and neither are plainly exposed through an api. Reasons for using blacklisted group in options:

- options API won't break the system
- only blackslists users by username not IP (might not be anticipated by user)
- allows user to un-ignore users easily (though config is clunky it is the main way users configure the app)
- probably what users expect if they think about it a second


jpdillingham:
This unfortunately touches on another sore area, which is the inability to update the configuration at run-time.

This, like UI config/preferences, is also blocking a lot of stuff, and there are no good solutions because most YAML libraries don't support "round tripping" (loading into memory, modifying, saving back to YAML) that preserves comments.  Because slskd leans heavily on the commented-out config it creates on startup as documentation, this would be a bad experience for a lot of users.

I have thought a lot about this over the years and the options are pretty much:

1. write my own/modify an existing YAML library that can preserve comments
2. add yet another layer of configuration that sits in a database

I don't really like either of these, which is why everything that would depend on this has been blocked

samhatter:
Ok so good news, in addition to the Document api, this yaml library also provides CST parsing which does preserve everything, not just semantically useful stuff (the document api would reformat whitespacing). This seems to work well for me, just a little more complex. Let me know what you think.

Referencing:
[CST Docs](https://eemeli.org/yaml/#working-with-cst-tokens)

samhatter:
Hi @jpdillingham,

Also getting back to this PR. I see I found a library that I believe does what we want wrt YAML parsing. Do you want me to assemble some sort of test suite to demonstrate various edge cases? That shouldn't be too tedious.

samhatter:
Ok I found a number of issues with the logic, so I cleaned it up and added a test suite that we can use to validate behavior and prevent regression. I also needed to update after the merge changed the options schema. I also added the ability to select which handlers are implemented on each page so in the future I can add a dm user feature that will only be present on rooms.

jpdillingham:
I'm going to take a stab at writing a C# library that can handle round-trip serialization of YAML; theoretically it should be straightforward if I don't worry about every edge case and just focus on the slskd options document.  If I can pull it off, I'll be able to add an API for updating options and a lot of things get unblocked

snapetech:
Closing this as part of a repo-wide PR reset before the next release. main has moved substantially since this was opened, so if we still want this change we should reopen it as a fresh PR against current main.

Issue #1425, opened 2025-08-18
mobile ui issue in portrait mode


jauling:
Landscape seems fine, I can see all the tabs under system.
<img width="2256" height="724" alt="Image" src="https://github.com/user-attachments/assets/7baa0f45-2ddf-48b4-b2ef-f9fe4ac94cea" />

But, once I use my mobile in portrait mode, the right most tabs disappear and unable to scroll to click them.
<img width="1080" height="709" alt="Image" src="https://github.com/user-attachments/assets/a598e529-3961-4e32-bfea-4cd1669f5c8e" />

jpdillingham:
I agree that it's a problem.  Responsive layouts in general could be much better.

I'm adding the help wanted tag because I won't have time to work on this any time soon.

tommytz:
I've added a small fix for this issue. Let me know if there are any problems.

PR #1517, opened 2025-11-09
fix: system menu scrolls when overflowing on narrow screens

tommytz:
This aims to fix https://github.com/slskd/slskd/issues/1425

The tab menu in the system page will now scroll when the the content overflows the container on the x-axis.

The overflow CSS was only added to the system menu. The header menu had issues with the right hand side menu disappearing when overflow was added to all elements with the `.menu` class (and it already scrolls nicely).

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .vs
 .vscode
+.idea/
 .DS_Store
 /src/slskd/bin/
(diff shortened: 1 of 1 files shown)


jpdillingham:
Looks great, thanks!

PR #1518, opened 2025-11-09
Add custom theming support

samhatter:
# Add custom theming support

## Description
 Adds the ability for theming though custom css file. A number of themes are provided in the docs. This is achieved by adding a configurable file that I can serve on endpoint and then load into the app.

## Changes
 - new /theme endpoint in the application controller to serve theme
- new `custom_css_path` configurable option
- three sample themes in `docs/themes`
- update to `docs/config.md` to explain how theming works

## Justification
- Theming is necessary if people are going to look at this in a variety of lighting conditions/monitors.
- This solution is simple and allows users to reuse themes.
- Later on we can even make config yaml that translates to css we serve to make it even more accessible.
- serving the css as a file instead of injecting the style through js (which I considered) is more sensible, performant and takes advantage of caching
- I included default themes because it wasn't trivial to get some workable examples and people can modify them to get what they want.


jpdillingham:
I'm definitely not opposed to the idea, but for something like this I would like to keep everything on the UI side.

slskd was created from the start to support different UIs, swappable by changing the path to the static content (or just overwriting it on disk) and I want to steer clear of anything that will too closely relate to one UI on the backend so things don't get confusing.

I think to fix this up all that is needed is to move the themes to the public directory in the UI and then store the theme somewhere in local storage

There's a lot of stuff blocked on UI options/preferences, and I think the easy answer there is to just leverage local storage with whatever approach makes sense.  It's not ideal that preferences won't be applied across devices but it's more ideal than any of the alternatives

samhatter:
Ok so like a user would just put the theme into the public directory or mount with docker. I like that actually. It's pretty jank to serve a stylesheet like this.

samhatter:
Ok I modified it so the user would just put the theme.css in their public dir. Pretty easy to just change where we load the file from and this solution serves it in a more optimized way.

samhatter:
Hi @jpdillingham,

Checking back in on this PR. I see that I have changed the architecture such that the backend doesn't serve the theme through API, the user just has to put it into the public dir to be served as a static asset. This adheres to your design principle of swappable UIs because the backend is unchanged and has no new responsibilities. Is there any further modification necessary or testing that you would like me to perform?

jpdillingham:
I will take a look this weekend.  The CSS files can be dropped in `/src/web/public` and should then be shipped along with the rest of the frontend, there just needs to be a way to choose which one to use.  These could be baked into the react app (I think) which would make selection easier (at least, easier to populate a dropdown).

samhatter:
> I will take a look this weekend.  The CSS files can be dropped in `/src/web/public` and should then be shipped along with the rest of the frontend, there just needs to be a way to choose which one to use.  These could be baked into the react app (I think) which would make selection easier (at least, easier to populate a dropdown). 

That makes sense. Currently it's hardcoded to theme.css in public (initializing defaults if not present), but a ui driven toggle could be a better solution for users that want to flip between styles. I'll give it some thought as well.

jpdillingham:
[2026-03-29-css-theme-switching-design.md](https://github.com/user-attachments/files/26335786/2026-03-29-css-theme-switching-design.md)
[2026-03-29-css-theme-switching.md](https://github.com/user-attachments/files/26335788/2026-03-29-css-theme-switching.md)

I took some time with the [superpowers](https://github.com/obra/superpowers) Claude plugin to flesh this idea out more and figured I spent the tokens so I might as well share the markdown it produced.  Curious your thoughts.

tl;dr this would put the theme css and a manifest in `/public/themes` and allow the user to choose between them on the fly (or supply their own custom CSS, stored in localStorage).

samhatter:
Sorry I forgot to attend to this message earlier. I read your markdowns, and I have some feedback on the design. I like the idea of a manifest and multiple shipped themes that can be toggled in react. I do think however that light/dark modes should be included with each theme file and that light/dark mode toggling should remain largely the same as it is now because I (and I assume others) like to toggle light/dark modes for different displays/TOD/lighting  whereas a theme is more of an aesthetic preference.

Aside from implementing some theme selector button, a lot of work to be done just in terms of making the front-end styling coherent enough to be easily adaptable and then creating some thoughtful default themes. I think I have some commits to that effect on my fork. I just took a vacation so maybe I'll submit a more polished concept soon.

PR #1520, opened 2025-11-10
Add support for changing user presence

### LINKED ISSUE #949: Change user presence status
Hey! How can I set my user status to `Away`?
(something like `SLSKD_SLSK_PRESENCE=away`?)

notsuu:
Implements a config option and API endpoint for setting the user's presence status. Also adds a button to the web UI for toggling said presence. Closes #949.

firefox_TBWZZMmJkl

Currently only does manual changes. Also I'm not too sure if dedicating an entire StateMonitor to a single UserPresence is a good solution but I don't know of any better ones.

### DIFF
--- a/src/web/src/components/App.jsx
+++ b/src/web/src/components/App.jsx
@@ -6,6 +6,7 @@ import * as relayAPI from '../lib/relay';
 import { connect, disconnect } from '../lib/server';
 import * as session from '../lib/session';
 import { isPassthroughEnabled } from '../lib/token';
+import { setPresence } from "../lib/users";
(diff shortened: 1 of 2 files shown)


jpdillingham:
Sorry I neglected to respond to this.  Right now a lot of things are blocked on the inability to update the configuration while the app is running (programmatically), so this would be difficult to implement as designed.  It could be done using the application state only and by considering a SignalR connection to a non-Relay-agent client to mean the user is active.

I want to be careful with this and not allow people to mark that they are active when they aren't, as it misrepresents how they are interacting with the network and the other users on it.

PR #1553, opened 2025-12-03
feat: Auto-replace stuck downloads with alternative sources

snapetech:
## Summary
Adds automatic detection and replacement of stuck/failed downloads with alternative sources from the Soulseek network.

## Features
- Detects downloads in stuck states (TimedOut, Errored, Rejected, Cancelled)
- Searches Soulseek network for alternative sources matching file extension
- Filters alternatives by configurable size threshold (default 5%)
- Ranks alternatives by size difference, free slots, queue length, upload speed
- UI toggle in Downloads page header
- API endpoints for manual and automatic replacement

## Configuration Options
```
--auto-replace-stuck          Enable auto-replacement (default: false)
--auto-replace-threshold      Max size diff % (default: 5.0)
--auto-replace-interval       Check interval in seconds (default: 60)
```

Or via environment variables:
```
AUTO_REPLACE_STUCK=true
AUTO_REPLACE_THRESHOLD=5.0
AUTO_REPLACE_INTERVAL=60
```

## API Endpoints
| Method | Endpoint | Description |
|--------|----------|-------------|
| GET | `/api/v0/transfers/downloads/stuck` | List stuck downloads |
| POST | `/api/v0/transfers/downloads/find-alternative` | Find alternatives for a download |
| POST | `/api/v0/transfers/downloads/replace` | Replace a stuck download |
| POST | `/api/v0/transfers/downloads/auto-replace` | Process all stuck downloads |
| GET | `/api/v0/transfers/downloads/auto-replace/status` | Get auto-replace config |

## UI Addition
The feature adds an "Auto-Replace" toggle to the Downloads page header.

## Testing
- Tested with multiple stuck downloads
- Verified alternatives are found within size threshold
- Confirmed successful replacement and enqueueing from alternative sources

### DIFF
--- a/src/slskd/Core/Options.cs
+++ b/src/slskd/Core/Options.cs
@@ -963,6 +963,32 @@ namespace slskd
                 [Description("the total download speed limit")]
                 [Range(1, int.MaxValue)]
                 public int SpeedLimit { get; init; } = int.MaxValue;
+
+                /// <summary>
+                ///     Gets a value indicating whether auto-replace for stuck downloads is enabled.
+                /// </summary>
+                [Argument(default, "auto-replace-stuck")]
+                [EnvironmentVariable("AUTO_REPLACE_STUCK")]
+                [Description("automatically replace stuck downloads with alternative sources")]
+                public bool AutoReplaceStuck { get; init; } = false;
+
+                /// <summary>
+                ///     Gets the maximum size difference percentage for auto-replacement.
+                /// </summary>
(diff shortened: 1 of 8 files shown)


jpdillingham:
Thanks for the effort, but functionality like this is complex and requires a much deeper strategy to work well for all but the simplest cases.

I'm not interested in features that solve only the easy parts because it increases maintenance pressure on me to fill in the gaps.

The functionality in this PR can be achieved using the existing API endpoints and an external script/application.

Issue #1083, opened 2024-02-19
Address binding features


melvyn2:
Here are a few (hopefully) minor features requests related to how slskd binds to listen addresses:
- Default listen addresses should include both IPv4 and IPv6 (e.g. `0.0.0.0, ::`)
- The ability to bind the web interface to a specific set of addresses instead of (currently) all IPv4 addresses
- The ability to bind to specific interfaces/devices instead of addresses, for both soulseek connections and the web server

jpdillingham:
The ability to bind to a specific interface address was added in #1018 and I neglected to reference this issue.  I missed a spot in the documentation and will close that gap shortly.

philippludwig:
But what about the web interface? Is `0.0.0.0` really hardcoded here?

jpdillingham:
It is at the moment.

christo2015:
I don't know if I should make a new request for this, but I would really like it to be possible to bind to both an IPV4 and IPV6 adress at the same time. Unless I am missing something, I can't bind both at the same time.

kittybwained:
I would also like this feature, I would like to be able to bind the web server to 127.0.0.1 so only I can access it without needing to set up firewalls.

x0vi:
This would be nice for my IPv6 only cluster, as i cant access it

jpdillingham:
Released in 0.24.2

PR #1555, opened 2025-12-03
feature: configurable ip address

x0vi:
(Based on the issue  #1083 )
This adds configurable IP address binding for HTTP/HTTPS listeners to support IPv6 and custom network interfaces for example.

### DIFF
--- a/src/web/package-lock.json
+++ b/src/web/package-lock.json
@@ -180,7 +180,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.23.9.tgz",
       "integrity": "sha512-5q0175NOjddqpvvzU+kDiSOAk4PfdO6FvwCWoQ6RO7rTzEe8vlo+4HVfcnAREhD4npMs0e9uZypjTwzZPCf/cw==",
       "dev": true,
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.23.5",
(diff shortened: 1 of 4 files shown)


jpdillingham:
I will pick this up and get it merged in a bit.  Some notes for myself:

The current logic uses `IPAddress.Any`, which is strictly an ipv4 address under the hood.  This should be `IPAddress.IPv6Any`, or the current `options.Listen()` call should be changed to `options.ListenAnyIP()` which appears to be the better of the two options:

```
    /// <summary>
    /// Listens on all IPs using IPv6 [::], or IPv4 0.0.0.0 if IPv6 is not supported.
    /// </summary>
    public void ListenAnyIP(int port, Action<ListenOptions> configure)
    {
        ArgumentNullException.ThrowIfNull(configure);

        var listenOptions = new AnyIPListenOptions(port);
        ApplyEndpointDefaults(listenOptions);
        configure(listenOptions);
        CodeBackedListenOptions.Add(listenOptions);
    }
```

If a user wanted to listen on just localhost the guidance would be to supply the proper loopback address for their system (`::1` or `127.0.0.1`), which should be reasonable if a user has a strong opinion about this.

The logic in the kestrel config is then:

If the user has specified any value for `ip_address`, use `options.Listen(<provided ip4/6 address>)`, otherwise use `options.ListenAny()`

The only uncovered use case would be instances where the user wanted to listen on multiple, specific addresses but not 'any', in which case another code/config change would be needed.  When CIDRs can be supplied for API keys or the relay, they can be supplied as a comma separated list.  I suppose this option should have that ability as well.

The option will need a custom validator that iterates over the supplied addresses to make sure they are valid.

PR #1600, opened 2026-01-14
Fix UrlEncodingModelBinder regression with PathBase

### LINKED ISSUE #1599: Regression in 0.24.2: API endpoints return 404 when running with base path
### Is there an existing issue for this?

- [x] I have searched for existing issues (both open and closed)

### Describe the Bug

When running slskd with a base path (e.g., `https://example.com/slskd/`), the "Search for Additional Files in This Directory" feature returns a 404 error for all search results. The API endpoint `POST /api/v0/users/{username}/directory` fails to find the route.

This is a regression introduced in version 0.24.2 and works correctly in version 0.24.1.

### Steps To Reproduce

1. Deploy slskd 0.24.2 with a base path configuration (e.g., behind a reverse proxy at `/slskd`)
2. Perform a search and get results from any user
3. Click "Search for Additional Files in This Directory" on any search result
4. Observe a 404 error in the browser console

Example failing request:
- URL: `https://media.example.com/slskd/api/v0/users/someuser/directory`
- Method: `POST`
- Body: `{"directory":"@@xxxx\\Artist - Album"}`
- Response: `404 Not Found`

### Environment

```markdown
- OS: Ubuntu 24.04
- slskd: 0.24.2
- Docker Install?: yes
- Filesystem: ext4
- Hardware: Lenovo ThinkCentre with Intel(R) Core(TM) i5-4570T
```

### Anything else?

Root Cause Analysis:

The issue was introduced in PR #1596 which added the `UrlEncoded` attribute to username parameters and modified the `UrlEncodingModelBinder` to use raw URLs to avoid double-decoding issues.

The bug occurs in `src/slskd/Common/Middleware/UrlEncodingModelBinder.cs`. The model binder calculates the parameter index based on the route template (which doesn't include the base path), but then applies that index to the raw URL (which does include the base path).
                                                                                                                                                                                                                                                                                                                            
Example:
- Route template: `api/v{version:apiVersion}/users/{username}/directory`
- Index of {username}: `3`
- Raw URL with base path: `/slskd/api/v0/users/someuser/directory`
- Element at index 3: "`users`" (should be "`someuser`")

andykelk:
## Description

Fixes a regression introduced in v0.24.2 where API endpoints return 404 errors when the application runs with a base path (e.g., `/slskd` via `UsePathBase`).

## Root Cause

The `UrlEncodingModelBinder` calculates the parameter index from the route template (which doesn't include the base path), but applies it to the raw URL (which does include the base path). This causes extraction of the wrong URL segment.

**Example:**
- Route template: `api/v{version:apiVersion}/users/{username}/directory`
- Index of `{username}`: 3
- Raw URL: `/slskd/api/v0/users/someuser/directory`
- Incorrectly extracts: `users` (should be `someuser`)

## Solution

Strip the `PathBase` from the absolute path before splitting into segments, ensuring correct parameter extraction regardless of base path configuration.

## Testing

- Added comprehensive unit tests covering scenarios with/without PathBase, query strings, and error handling
- All tests passing
- Verified fix resolves the reported issue

## Related

Fixes #1599

### DIFF
--- a/src/slskd/Common/Middleware/UrlEncodingModelBinder.cs
+++ b/src/slskd/Common/Middleware/UrlEncodingModelBinder.cs
@@ -76,7 +76,17 @@ public class UrlEncodingModelBinder : IModelBinder
             var rawUrl = bindingContext.HttpContext.Features.Get<IHttpRequestFeature>()?.RawTarget ?? string.Empty;
 
             var request = bindingContext.HttpContext.Request;
-            var rawValue = new Uri($"{request.Scheme}://{request.Host}{rawUrl}").AbsolutePath // discard query string
+            var absolutePath = new Uri($"{request.Scheme}://{request.Host}{rawUrl}").AbsolutePath; // discard query string
+
+            // if the application is running with a base path (e.g., /slskd), we need to remove it
+            // from the absolute path before splitting, as the route template doesn't include it
+            var pathBase = request.PathBase.Value ?? string.Empty;
+            if (!string.IsNullOrEmpty(pathBase) && absolutePath.StartsWith(pathBase, StringComparison.OrdinalIgnoreCase))
+            {
+                absolutePath = absolutePath.Substring(pathBase.Length);
+            }
+
+            var rawValue = absolutePath
                 .Split('/', StringSplitOptions.RemoveEmptyEntries)
                 .ElementAtOrDefault(index);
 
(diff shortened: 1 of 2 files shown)


jpdillingham:
Thanks for reporting this and supplying the fix.  The solution works, I'll merge it and tweak it a bit to taste.

PR #1604, opened 2026-01-15
Preventing Safari Autofill in Message Input Elements

samhatter:
I've noticed that sometimes safari will try and autocomplete information within room message input element. I've experienced safari attempt to autofill email, password, even credit card information. This is dangerous because it could lead to user's browsers inadvertently doxxing them. I've read some information online indicating that safari ignores the id attributes on input elements but gives some attention to the name attribute to I'm going to try that.


samhatter:
It seems to work ok but I'll test it out a bit more to see if Safari has truly decided to get outta my business and then submit pr

jpdillingham:
What do you think about using `autocomplete=off`? docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

samhatter:
That's what one would expect right!? Turns out we already on have it on the element- safari is just incredibly persistent and actually this setup doesn't work 100%

samhatter:
I changed to a different strat of making the id less ambiguous, no name, and adding a placeholder. Hope this helps, will confirm in a while if I stop noticing this behavior since it is rather non-deterministic.

jpdillingham:
@snapetech you've added this comment to every open PR in this repo.  Can you please remove them, and stop doing it?

jpdillingham:
@snapetech and please leave a message explaining the mixup in any PR that's not owned by me.  These folks all got this message and it won't be clear that you aren't a maintainer.

snapetech:
I've deleted all of the accidental posts, including my last reply here so adding again;

Apologies for the noise.  I had an agent set to deal with dependabot PR's and while it did that, it then turned around and started trying to comment and close on the upstream project PR's instead of my own.

I've deleted all of those posts, feel free to point anyone to this post -- my mistake, I've added additional guardrails to that agent to (try to) prevent this from happening again.

samhatter:
i am no longer a safari user.

PR #1632, opened 2026-02-08
Add VPN integration and gluetun client

jpdillingham:
This PR builds on #1623, adding a new VPN integration that polls external VPN clients for status updates and controls slskd accordingly.

The only VPN client currently supported is [Gluetun](https://github.com/qdm12/gluetun), but others can be added in the future if they support the ability to poll for status.

Options:

```yaml
integration:
  vpn:
    enabled: false
    port_forwarding: false
    polling_interval: 2500
    gluetun:
      url: http://localhost:8000
      timeout: 1000
      auth: apikey # or basic or none
      api_key: <gluetun API key>
      username: <username if using basic auth>
      password: <password if using basic auth>
```

When the VPN integration is enabled, status polling begins just before the client attempts to start for the first time.  The connection will wait until the VPN is "ready", which means that it has indicated that it is connected, and if `port_forwarding` is enabled, has supplied the forwarded port.

If the VPN client goes down while slskd is running, slskd will disconnect (within `polling_interval`) and wait for the VPN to come back up before reconnecting.  Note that this is NOT a substitute for routing all slskd traffic through the VPN; that is the responsibility of the user.  slskd waits to connect and disconnect so that the Soulseek server has an accurate IP address and port.

The application state has also been updated to provide VPN information, and feedback about whether the application is waiting to connect because of the VPN:

image

When the VPN is disconnected:

image

## Tests

* [x] ~~If `required`, wait to connect until VPN is up (and we have a port, if forwarding is on)~~ removed required option; if VPN is enabled it is required
* [x] If `required`, disconnect if the VPN goes down or polling fails
* [x] Connects properly and applies forwarded port when starting with VPN up
* [x] Connects properly and applies forwarded port when starting with VPN down
* [x] Waits for forwarded port when port forwarding is enabled
* [x] Does not wait for forwarded port when port forwarding is disabled
* [x] Respects polling interval option
* [x] ~~If not `required`, stop VPN service polling when connection watchdog stops (we don't need to react to disconnects)~~ It needs to keep polling to update IP and location, etc
* [x] Option validation works properly
* [x] State updates work properly (and make sense)

### Auth

* [x] API key
* [x] Basic

### Validation

* [x] PollingInterval
* [x] Version
* [x] Url
* [x] Timeout
* [x] Auth
* [x] Username/password/api key

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1168,6 +1168,53 @@ integration:
 
 System-defined integrations are part of the application logic.  Users can configure settings, but don't have any control over behavior.
 
+### VPN
+
+> [!CAUTION]
+> **Configuring a VPN integration does not change how network traffic to the Soulseek server or other peers is routed**.
+> 
+> Users wishing to use a VPN are responsible for ensuring slskd is connecting through their VPN's network interface and therefore routing traffic through the VPN.  This integration simply provides visibility; a way to easily apply dynamically forwarded ports, and some extremely light protection for users who do not use a "kill switch" to drop internet connectivity entirely if their VPN client disconnects.
+
+The VPN integration can poll an external service/application for VPN status (connectivity, forwarded port, etc) and connect or disconnect the connection to the Soulseek server according to status.
+
+Upon first start slskd will begin polling the configured VPN and will delay connection to the Soulseek server until the VPN returns a status that indicates it is connected.  If `port_forwarding` is enabled, the VPN must also provide the forwarded port for slskd to consider it "ready".
+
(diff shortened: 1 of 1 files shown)


jpdillingham:
There seems to be an issue with trying to apply the overlay before things are fully up: 

```
[slskd.Application] [18:29:28 ERR] Failed to apply option update: Failed to start listening on 0.0.0.0:59782; the IP and/or port may be in use or are otherwise unavailable
Soulseek.ListenException: Failed to start listening on 0.0.0.0:59782; the IP and/or port may be in use or are otherwise unavailable
```

Issue #1152, opened 2024-06-20
The 'Search Again' button doesn't work


jpdillingham:
This is somewhat of a meta-issue to address a number of issues.  At some point along the way it appears to have broken entirely.

Additional issues:

#1087 
#918

U99c:
it's not just "search again" for me - it seems to be a problem for all forms of search initiated from the search page, as neither can i get full directory contents.  instead, i get the error: "the wait timed out after 5000 milliseconds"

this IMO is a significant problem.  for example, you may be hunting for an artist or album with a difficult search term, so you use a song name to refine the results - but without the ability to get the full directory contents, it is impossible to download the desired collection or album using slskd'd UI.  

in slskd's current state, its usability is greatly diminished by this bug.  i would not consider this backlog material.

jpdillingham:
This was fixed a while ago

PR #1666, opened 2026-03-09
Fix "Search Again"

rfletcher:
As described in issue #1152, the "Search Again" button isn't working. This change fixes the button for me locally, adding a missing `search` key to a function call.

The bug appears to have been introduced as part of this large `eslint` commit, where the `search` key was added in one place, but omitted from another: https://github.com/slskd/slskd/commit/4bcbde71c847a7a8c73d8ff2399a84f1d61fe766 (Scroll to `SearchDetail.jsx`, lines 130-132. I tried linking directly, but GitHub is having trouble due to the size of the commit.)


jpdillingham:
Thanks, it's working locally for me too.

PR #1696, opened 2026-04-11
fix favicon.ico relative path

deepsweet:
In case slskd is hosted not under `/` base URL but rather something like `/slskd`.

By default browsers try to fetch `/favicon.ico` which is not there, but this PR fixes it to become relative to the HTML file.


jpdillingham:
Thanks!

PR #1698, opened 2026-04-12
Add errors for deprecated keys

jpdillingham:
#1672 reorganized some keys in the configuration to make room for upcoming features, but because the startup validation permits keys in the YAML that don't match properties in the `Options` class, this meant that users who don't read release notes wouldn't know this had happened until they realized that their settings weren't being respected.

This PR adds these deprecated properties back to `Options` and causes validation to fail explicitly if they are present.

### DIFF
--- a/src/slskd/Core/Options.cs
+++ b/src/slskd/Core/Options.cs
@@ -253,6 +253,12 @@ namespace slskd
         [Validate]
         public SharesOptions Shares { get; init; } = new SharesOptions();
 
+        [Obsolete("temporary sentinel to warn users about breaking change.  use Transfers instead.")]
+        public object Global { get; init; } = null;
+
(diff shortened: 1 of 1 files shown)


onyxiarivera:
Why couldn't this have been automatically updated? I haven't needed my instance in a bit and I come back to it spitting out logs every 500ms for days on end.

jpdillingham:
I'm using semver, and the major version of 0 means that things may change significantly.  Expect to need to read the release notes and maybe make some adjustments when the minor version changes (e.g. 0.24 -> 0.25).

More specifically:

1. The application modifying its own config file doesn't seem quite right to me
2. There is no yaml library for C# that can support round-trip editing of files without losing formatting and comments
3. I do this for fun, in my limited free time.  I don't think it is too much to ask for people that have gotten value out of slskd to do a little bit of work so that they can take advantage of the new features and functionality I'm giving them

PR #1699, opened 2026-04-13
feat: add username reg pattern matching to blacklist group

ofernander:
Adds a  field to  that accepts
a list of regular expressions matched against usernames.

AI disclosure 

Claude was used to review slskd code and provide info to me needed for changes, then used to validate and correct my changes. 

Changes were interactive and tested, code was not just vibe coded without intention.

### DIFF
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -112,6 +112,8 @@
 #     blacklisted:
 #       members:
 #         - <username to blacklist>
+#       patterns:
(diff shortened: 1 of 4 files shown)


jpdillingham:
See #1702 for a discussion about this

jpdillingham:
I'm going to fix this up a bit before I merge into master; I want to extend this capability to the leechers group and need to compile the patterns after updates

PR #1754, opened 2026-06-04
Add Traffic & File Tickers in Web UI

### LINKED ISSUE #710: Add Traffic & File Tickers in Web UI
### Discussed in https://github.com/slskd/slskd/discussions/547

<div type='discussions-op-text'>

<sup>Originally posted by **ckbaudio** July 14, 2022</sup>
Would be nice to have a total traffic and file ticker within the 'shares' option or somewhere else in the UI.
I'm currently using NetHogs for this but would be nice to have it in the WebUI :)

image
</div>

TheMagzuz:
Closes #710

this implementation is maybe a bit hacky, as it uses the `/telemetry/reports/transfers/summary` endpoint, which turns out to have all the information needed to replicate the UI in the issue.
the endpoint doesn't expose the date of the oldest record however, so that isn't included currently.
i see two approaches to tackle this, either
1. extend `TransferSummary` with a `EarliestRecordTime` property, or
2. make a dedicated endpoint for traffic data.

i personally think the second approach is cleaner, since it appears that `TransferSummary` is produced in multiple places, but i'd like your input on that.
in case of the second approach, we also need to figure out where to put the endpoint, since it doesn't really feel like it fits nicely with any of the current endpoints (`/telemetry` seems to be for public endpoints, `transfers` could maybe work?)

i'm also not a huge fan of placing it under `/system/shares` in the UI, since it seems quite unrelated to the other things on that page.
maybe in the future there should be a dedicated page for showing stats (maybe making `/system/info` look more user-friendly, since it seems to already show a lot of interesting statistics)

### DIFF
--- a/src/web/src/components/System/Shares/TrafficStatsTable.jsx
+++ b/src/web/src/components/System/Shares/TrafficStatsTable.jsx
@@ -1,0 +1,66 @@
+import { Table } from 'semantic-ui-react';
+
+const formatBytes = (bytes, decimals = 2) => {
+  if (!Number(bytes)) return '0 Bytes';
+
+  const k = 1_024;
+  const dm = Math.max(0, decimals);
+  const sizes = [
+    'Bytes',
+    'KiB',
+    'MiB',
+    'GiB',
+    'TiB',
+    'PiB',
+    'EiB',
+    'ZiB',
+    'YiB',
+  ];
+
+  const index = Math.floor(Math.log(bytes) / Math.log(k));
+  const scaled = (bytes / k ** index).toFixed(dm);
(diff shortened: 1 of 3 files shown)


jpdillingham:
Thanks for the PR, unfortunately I've already got a branch going where I'll be adding the same stuff.  Some high level KPIs will be made available via a SignalR hub (so, real time) and the rest can be polled from various endpoints.

If you have an interface that you like, your effort isn't lost; the next thing I'll be working on is a way for users to swap the entire UI out with one of their choosing.

jpdillingham:
This was implemented in #1778, thanks for the contribution, though!

PR #1756, opened 2026-06-06
Move 'permissions' options under 'destination', improve application of directory mode

jpdillingham:
When I originally added the option to explicitly set the unix file mode of downloaded files and directories, I used a method that causes the effective mode to be influenced by the application's umask.  This meant that if someone specified `777`, the file would have the desired permissions but the directory would end up with something else, typically `755` (`022` being a common default umask)

I recently realized that I can explicitly set the mode of directories as well as files by calling `File.SetUnixFileMode`, which is pretty poorly named.  Though on unix, directories are files, too.  So _technically_ it's accurate.  Anyway..

After I added the download destination options in #1746, that group of options became a much better home for this option than in the top-level `permission` key, so I'm moving it.

As with other recently moved/deprecated keys, I am adding a fake validation error to alert users to the move.  In addition to moving the option, the behavior is changed and users need to reconsider their desired mode, knowing that it now correctly applies to directories as well as files.

Docs:

### Permissions

On [Unix-like](https://en.wikipedia.org/wiki/Unix-like) operating systems, downloaded files will be created according to the [umask](https://en.wikipedia.org/wiki/Umask) of the process (see [Permissions](#permissions) for more info).

Users can optionally configure alternative permissions by setting the permission mode option which, when enabled, effectively causes slskd to `chmod` files and directories with the configured mode immediately after creation.

This option has no effect on Windows.

#### **YAML**
```yaml
transfers:
  download:
    destination:
      permissions:
        mode: 644 # chmod syntax, e.g. 644, 777.  has no effect on Windows
```

## Testing

* [x] Validation error when old permissions key is used
* [x] Files and directories are created with the expected mode

### DIFF
--- a/config/slskd.example.yml
+++ b/config/slskd.example.yml
@@ -81,6 +81,8 @@
 #     destination:
 #       subdirectory: ${SOURCE_DIRECTORY} # options: SOURCE_USERNAME, SOURCE_PATH, SOURCE_DIRECTORY, BATCH_ID, BATCH_EXTERNAL_ID, SEARCH_ID, SEARCH_TEXT
 #       exists: rename # 'overwrite' or 'rename'
+#       permissions:
+#         mode: ~ # chmod syntax, e.g. 644, 777. does nothing on Windows. 
 #   groups:
 #     default:
 #       upload:
--- a/docs/config.md
+++ b/docs/config.md
@@ -117,21 +117,11 @@ Users wishing to create files with different permissions can change the umask va
 
 When running within a docker container the environment variable `SLSKD_UMASK` can be used to change the umask for the process within the container, and this will in turn change the resulting file mode on a mounted volume.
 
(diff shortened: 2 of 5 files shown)


meatsnails:
why does the config have to have breaking changes every version

jpdillingham:
@meatsnails I'll make sure I include one in all future releases just for you

meatsnails:
> @meatsnails I'll make sure I include one in all future releases just for you

Still didn't awnser the question

jpdillingham:
The explanation for this specific change is in the description.

The explanation for why there are "breaking changes every version" is because slskd has not reached version 1.0.0, and 0.* releases under [semver](https://semver.org/) are meant _specifically_ to communicate that breaking changes [may occur at any time](https://semver.org/#spec-item-4):

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

I increment the minor version when substantial or breaking changes are released.  If you don't want to experience disruption when you upgrade, you should read the release notes of those before upgrading.

PR #1758, opened 2026-06-07
Change all the scripts in bin/ to be more OS agnostic.

breakone9r:
Literally all that has been done is replace the bash shebang (The #!/interpreter line) to be more cognizant of where various other operating systems install bash, rather than hard coding a path.

`#!/bin/bash` has been changed to `#!/usr/bin/env bash` so that as long as bash is in your $PATH$ it will still exectute the script.

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 web_only=false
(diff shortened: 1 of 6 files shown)


jpdillingham:
Thanks! One small typo of a missing slash in the publish script; I'll merge this then patch it directly to master

PR #1775, opened 2026-06-26
stricter and configurable cors

samhatter:
CORS settings are ridiculously loose. I honestly don't think it matters *that much* right now because we store the JWT in local storage and not like, say a cookie with none same-site policy. That is sorta its own can of worms because it could make an XSS attack worse, but closes the potential CSRF. I made copilot write this up to allow users to still configure CORS if they are using it.


jpdillingham:
Thanks for suggesting this, I could see it being useful.  To avoid any potential disruption to existing installs this will need to have the same (wide open) configuration by default, then users can opt in and tweak things if they want.

I think this would probably be:

```yaml
   cors:
     enabled: false # allow * 
     allowed_origins:
       - https://app.example.com
     allow_credentials: false
```

samhatter:
I think it is absolutely not worth preserving the allow_credentials setting defaults as there is no existing auth model that can use allow_credentials. Honestly maybe we should delete this altogether.  This could enable a CSRF in the future if we ever decide to harden ur auth and move to a cookie based jwt (and critically mess up and leave samesite=None) but as it stands we do not store any credentials anywhere a CORS request could conceivably touch (local storage is same-origin) so i strongly recommend we leave allow_credentials defaulting to false or just take that option out. If we want to support wild card cors hosts then this might be good practice since browsers disregard wildcard host + allow credentials, hence ur little middleware hack.

I would also hesitate to keep CORS wildcard by default. I'd imagine the amount of users who have slskd api calls embedded in other webpages is rather small to begin with, and it's not like breaking configuration changes are rare in this project. Additionally it is a real security vulnerability, if a little niche.  Consider unauthenticated setups behind some VPN or LAN. If an attacker knows the internally resolving url of a victim's  webui then an api request to that url would resolve and can be read by JS and exfiltrated because of the CORS setting (without which the requests could be made but like the contents wouldn't be readable so its stilll p bad even w proper CORS). 

TLDR; preserving allow credentials default is potentially insecure to future changes and is not relied on by any areas of the codebase. Preserving cors wildcard is sort of a bad idea imo and his some implications in unauthenticated intranet-only web-uis, but could actually be relied on users who embed slskd apis in other sites.

LMK what u think. Am I missing something?

jpdillingham:
I'm still thinking about the exact implementation details on this one so I'm going to hold off on including it in this next release.

There are several apps that run in their own docker container and make requests to slskd from the browser, and I don't have a good feel for how many people are using those apps.  If I remove the current configuration that's effectively `*` those apps will break and I don't really want the blowback.

If I do end up disabling the current pass-all by default I want to release it by itself to give the best chance of impacted users finding it instead of creating an issue.

I don't really have a good way of communicating breaking changes to users other than preventing the app from starting up.  In previous cases this was warranted because whatever was changing had the potential to expose them to something they may have opted out of; in this situation it would be to warn them about a breaking change that most likely wouldn't impact them.

PR #1779, opened 2026-07-07
Add recursive folder download support, closes #807

### LINKED ISSUE #807: How to download all folders in a folder?

SankeerthNara:

### DIFF
--- a/src/web/src/components/Browse/Browse.jsx
+++ b/src/web/src/components/Browse/Browse.jsx
@@ -220,10 +220,27 @@ class Browse extends Component {
     };
   };
 
-  selectDirectory = (directory) => {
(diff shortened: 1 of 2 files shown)


jpdillingham:
Closed for not following contribution guidelines

SankeerthNara:
Sir i completed the code where i added a button to download multiple files
can you go through it

PR #1786, opened 2026-07-11
fix: Set Cache-Control: no-cache on index.html responses

### LINKED ISSUE #1785: Stale index.html cache causes API failures after upgrade
### Is there an existing issue for this?

- [x] I have searched for existing issues (both open and closed)

### Describe the Bug

After upgrading slskd, the browser can serve a cached index.html that references old hashed JS filenames, causing API calls to fail because the endpoints they reference no longer exist. A refresh fixes it.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Linux
- slskd: 0.25.1
- Docker Install?: No 
- Filesystem: ext4
- Hardware: Raspberry Pi 3 Model B Plus Rev 1.3
```

### Anything else?

_No response_

apkatsikas:
Closes https://github.com/slskd/slskd/issues/1785

Verified localhost response headers (`https://localhost:5031/`):

image

### DIFF
--- a/src/slskd/Program.cs
+++ b/src/slskd/Program.cs
@@ -63,6 +63,7 @@ namespace slskd
     using Microsoft.Extensions.FileProviders;
     using Microsoft.Extensions.FileProviders.Physical;
     using Microsoft.IdentityModel.Tokens;
+    using Microsoft.Net.Http.Headers;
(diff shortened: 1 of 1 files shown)


jpdillingham:
Thanks!

PR #1787, opened 2026-07-12
Add recursive folder download support, closes #807

### LINKED ISSUE #807: How to download all folders in a folder?

SankeerthNara:
**Title:** Support recursive folder downloads from the browse interface

**Description:**

Closes #807

### Problem
Currently, downloading a folder that contains only subfolders (no files directly inside it) doesn't work — the browse download logic doesn't recurse into nested directories.

### Solution
Implemented recursive folder download support in the backend download logic so that nested subfolders and their files are included when a folder download is initiated.

### Testing
- [x] Downloaded a folder containing only subfolders and confirmed all nested files are queued
- [x] Downloaded a folder with a mix of files and subfolders and confirmed both are handled correctly
- [x] Verified existing (non-recursive) folder downloads still work as before

### AI Disclosure
I used Claude (Anthropic) to assist with writing portions of the backend code for this change. I've reviewed, tested, and take responsibility for all code submitted in this PR.

### DIFF
--- a/src/web/src/components/Browse/Browse.jsx
+++ b/src/web/src/components/Browse/Browse.jsx
@@ -220,10 +220,27 @@ class Browse extends Component {
     };
   };
 
-  selectDirectory = (directory) => {
-    this.setState({ selectedDirectory: { ...directory, children: [] } }, () =>
-      this.saveState(),
+  // Recursively walks a directory node and all of its descendant
+  // directories, flattening every file into a single array with
+  // fully-qualified (path-prefixed) filenames. Used to support
+  // "download folder including subfolders".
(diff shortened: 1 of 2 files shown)


jpdillingham:
Thanks for the contribution, but this conflicts with work I've been doing for the past several days in #1788

PR #1838, opened 2026-09-06
fix favicon.ico relative path in footer

deepsweet:
Hi.

Same as my previous PR #1696 but this time in footer.

In case slskd is hosted not under `/` base URL but rather something like `/slskd`.


jpdillingham:
Thanks

